PR #65, opened 2026-07-30
ci: fetch full history before moving the release branch

croffasia:
The `Move the release branch` step failed with:

```
! [rejected]        HEAD -> release (fetch first)
```

`actions/checkout` clones with `fetch-depth: 1`, so the commit the remote `release` branch points at is not present locally. git cannot prove the update is a fast-forward and rejects the push client-side, even though it is one.

`fetch-depth: 0` gives the step the full history.

`release` has been moved to v0.5.0 manually, since re-running the failed job would not cut a release again.


croffasia:
recheck

PR #96, opened 2026-08-04
feat: add checklists to board cards

### LINKED ISSUE #94: [feature]: Add Checklists to Board Cards
### Problem

Board cards already support subtasks, but not every task needs to be split into separate subtasks. In many cases, users simply need a lightweight checklist to keep track of small steps, reminders, or verification items without the overhead of creating and managing subtasks. Today, users have to either create unnecessary subtasks or maintain manual lists in the card description, making it harder to track progress.

### Proposed behaviour

As a user, I want every board card to have a Checklists section where I can create one or more checklists. Each checklist should have a title and contain multiple checkbox items that I can add, edit, reorder, check off, and delete. This gives me a quick way to track simple to-do items within a card while keeping subtasks reserved for larger, independent pieces of work.

_No response_

### Contribution

ahmedosamaft:
## What

Every issue can hold checklists of small steps that do not warrant subtasks of
their own. A checklist has a title and checkbox items that can be added, edited,
checked off, reordered and deleted; both the checklists of an issue and the items
of a checklist are dragged into order.

Checklists come back with the issue detail read, so the panel renders without a
second request. Adding, renaming and removing a checklist or an item is written to
the activity feed; checking a box is not, since a toggle is the most frequent write
here and would bury the rest of it.

## Why

A subtask is a full issue: it takes a sequence number, a column and a row on the
board. Small verification steps and reminders do not need that, so they end up
either as throwaway subtasks or as a manual list in the description, where nothing
tracks progress.

Closes #94

## How to test

1. `bun run db:migrate`, then `bun run dev`.
2. Open an issue, expand **Checklists**, and add a checklist and a few items.
3. Check one off — the heading tally moves immediately and survives a reload.
4. Drag an item to another position, and drag one checklist above another; reload
   and both orders persist.
5. Rename a checklist and an item inline, delete an item, then delete the whole
   checklist and confirm its items went with it.
6. Open the activity feed: `added checklist`, `added … to …` and the removals are
   there, and checking a box adds no entry.
7. Repeat in the board slide-over, the full issue page and the inbox split view —
   one mount point serves all three.
8. As a member without `work_items:edit`, the section is read-only: no add, delete
   or drag handles.

`bun --filter='api' run test` covers the API side (23 checklist tests).

## Checklist

- [ ] New environment variables documented in `.env.example` — none added
- [ ] Docs updated (`README.md` or the relevant `AGENTS.md`) — no rule changes needed

## Screenshots

image

### DIFF
--- a/apps/api/src/issues/__tests__/integration/checklists.test.ts
+++ b/apps/api/src/issues/__tests__/integration/checklists.test.ts
@@ -1,0 +1,360 @@
+import { describe, it, expect, beforeEach } from 'bun:test';
+import { authedApi, type Api } from '../../../__tests__/helpers/app';
+import { signUpTestUser } from '../../../__tests__/helpers/auth';
+import { resetDb } from '../../../__tests__/helpers/db';
+
+// Checklists: an issue holds several checklists, each holding checkbox items. Both
+// are ordered by position within their parent and come back with the issue read
+// (GET /issues/:issueId → `checklists`). A checklist and its items are part of the
+// issue, so they take the same work_items permission. The structural changes are
+// written to the activity feed; checking a box is not.
+
+interface Setup {
+  asOwner: Api;
+  columnId: number;
+}
+
+async function setupProject(): Promise<Setup> {
+  const owner = await signUpTestUser();
+  const asOwner = authedApi(owner.cookie);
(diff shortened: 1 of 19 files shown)


ahmedosamaft:

croffasia (approved):

PR #146, opened 2026-08-14
feat: add WIP limits to board columns

### LINKED ISSUE #127: [feature]:  Column (lane) limits on the kanban board — a configurable soft or hard maximum
### Problem

Kanban columns show how many issues they hold, but there's no way to say how many they *should* hold. Work-in-progress limits — the core mechanic of kanban — have to be tracked in someone's head or in a separate doc. "In Progress" quietly grows to fifteen cards; the counter reports the number faithfully, but a bare number carries no judgement, so nobody notices until standup. The board gives no feedback at the moment the limit is crossed, which is when it would actually change behaviour.

Teams also disagree on how strict a limit should be. Some want a nudge they can override when reality demands it; others want the limit to actually hold, because a limit that is always exceeded stops meaning anything. One fixed behaviour won't serve both.

### Proposed behaviour

- Allow an optional **maximum** to be set per column. When set, the existing counter renders as `count / max` instead of a bare count; columns with no maximum are unchanged.
- Alongside the maximum, a per-column **enforcement mode**, defaulting to soft:
  - **Soft (default).** Dragging a card into a full column still works — nothing is blocked, no confirmation dialog. The signal is visual only: at or over the limit the counter switches to a warning treatment (e.g. red text, and a subtle tint on the column header), and returns to normal when the count drops back under.
  - **Hard.** A drop that would take the column past its maximum is rejected. The drag shows the column as an invalid target while it's over capacity, the card animates back to where it came from, and a short toast explains why, naming the column and its limit. Same visual treatment as soft mode for a column already at the limit.
- Hard mode applies wherever an issue enters the column, not just drag-and-drop: the state picker on the issue detail, quick actions, and bulk state changes are all blocked with the same message, so the limit can't be bypassed by using a different control.
- Ways a hard limit can legitimately be exceeded, which the UI must handle rather than deadlock on:
  - Lowering a maximum below the current count, or a bulk import, leaves the column over capacity. This is allowed — the column shows the warning state, blocks further additions, and stays that way until it drains.
  - A per-project permission (e.g. project admins, or a dedicated "override WIP limits" permission) that turns a hard rejection into a confirmation dialog with a reason field, recorded in the issue activity. Teams that don't want an escape hatch simply grant it to nobody.
  - API and agent writes: see below.
- Decide explicitly whether the limit counts *all* issues in the column or only those matching the active filters. Filtered counts make the limit inconsistent between viewers, so the limit should evaluate against the unfiltered column total, with the header showing the filtered count when a filter is active and the limit still measured against the real one.
- Optionally a second threshold for "approaching the limit" (e.g. amber at max − 1), so the board warns before the limit is crossed rather than after.
- Configuration lives with the board/state settings, next to where states are defined, and is per project.
- Expose both the limit and the enforcement mode through the REST API and MCP alongside the state definition, so they can be set and read programmatically, and include them in state-related webhook payloads. A write that would breach a hard limit returns a `409` with a machine-readable error code naming the column and its maximum, so agents and integrations can react rather than fail opaquely — this matters because agents move issues across states without a human watching. An explicit override flag on the request, subject to the same permission as the UI, allows a deliberate breach.
- If two-level grouping is active, the limit applies to the column as a whole, with the indicator in the column header; per-swimlane counts stay plain.

_No response_

### Contribution

ahmedosamaft:
## What

A column can carry a maximum number of issues, with a mode deciding what happens
at it. Soft only warns: the header counter reads `count / max` and turns red at or
over the limit. Hard refuses an issue entering a full column, from every control —
drag and drop, the state picker, the context menu, the command palette, bulk
actions, the REST API and agents — since the check sits in the one store function
they all write through.

The count a limit is measured against is the column's real occupancy, not the cards
left by the active filters, so the limit means the same thing to every viewer.
Where filters hide part of a column, the header shows both numbers.

A refused write returns 409 with the code `wip_limit_exceeded`, which is what an
agent moving issues without a human watching can react to. Carrying a code needed
one on `HttpError` and an optional `code` on the shared error body; every existing
failure leaves it unset.

Ways a column legitimately ends up over its limit are allowed rather than
deadlocked: a limit lowered under the current count keeps its issues and takes
nothing new, and deleting a column still moves its issues out. Automations are not
the ones refused either — subtask rules and the GitHub handler run after the write
that triggered them, so they leave an issue where it is instead of failing an
action that already succeeded.

Limits are set per state in the project's state settings, and both fields are
exposed through the REST API and the MCP column tools.

## Why

Kanban columns show how many issues they hold, but there is no way to say how many
they should hold. Work-in-progress limits have to be tracked outside the tool: "In
Progress" grows to fifteen cards and the counter reports that faithfully, but a
bare number carries no judgement, so nothing pushes back at the moment the limit is
crossed.

Teams also disagree on how strict a limit should be — some want a nudge they can
ignore, others want the limit to hold — which is why the limit carries a mode
rather than one fixed behaviour.

Closes #127

## How to test

1. `bun run db:migrate`, then `bun run dev`.
2. Settings → States: set "In Progress" to max 3, leave the mode on Warn. The board
   header reads `count / 3`.
3. Drag a fourth card in — it lands, and the count plus the header tint turn red.
   Drag one out and it returns to normal.
4. Switch the mode to Block. The column refuses the fourth card during the drag and
   a toast names the column and its limit.
5. Try the same move from the issue detail state picker, a right-click on a card,
   the command palette, and a multi-select bulk action: each is refused with the
   same message.
6. `curl -X PATCH .../issues/<id> -d '{"columnId":<full column>}'` → 409 with
   `code: "wip_limit_exceeded"`. Same through an agent's `update_issue`.
7. Filter the board so most of the column is hidden: the limit still measures the
   real total and the header shows both counts.
8. Lower the max below the current count — allowed; the column keeps its issues,
   shows the warning state and takes nothing new.
9. Group the board by assignee: no limit badges. Group by status with swimlanes on:
   the badge is on the column header, per-swimlane counts stay plain.
10. Reorder a card within a full Block column — still allowed.

`bun --filter='api' run test` covers the API side (19 WIP tests).

## Checklist

- [ ] New environment variables documented in `.env.example` — none added
- [ ] Docs updated (`README.md` or the relevant `AGENTS.md`) — no rule changes needed


croffasia:
image

I suggest we keep the current color for blocked, and use something closer to yellow for warning.

croffasia:
One thing before this can go in: the new UI strings are English-only.  Please add the same keys to the remaining locales

ahmedosamaft:
> image
> I suggest we keep the current color for blocked, and use something closer to yellow for warning.

This one is for warning
 image

and this for the blocked 
image

ahmedosamaft:
> One thing before this can go in: the new UI strings are English-only. Please add the same keys to the remaining locales

Fixed in the last commit ✅

croffasia:
Thanks for the contribution! Merged

PR #150, opened 2026-08-14
feat: add Arabic language support with RTL layout

### LINKED ISSUE #126: [feature]: Arabic language support (i18n + RTL layout)
### Problem

The interface is English-only and laid out left-to-right, with no way to switch language. For Arabic-speaking teams this makes the product hard to adopt: every label, menu, empty state, and system message is in a second language, and the layout runs against the direction people read. Mixed Arabic and English content — an Arabic issue title next to an English label, or an Arabic sentence containing a Latin identifier like `ITSAPLAN-42` — renders with the wrong base direction, so punctuation and numbers land on the wrong side and titles read as garbled even though the stored text is correct.

There is also no separation between UI strings and code today, so this isn't just "add Arabic" — nothing is translatable at all. Any second language, Arabic or otherwise, needs that groundwork first.

### Proposed behaviour

- Extract all UI strings into translation files (one namespace per app area), with English as the source locale, and wire up an i18n layer in the Next.js app. This is the prerequisite; Arabic is the first locale that proves it works.
- Add Arabic (`ar`) as a locale, with a language selector in user settings and an instance-level default. The choice persists per user and is respected in server-rendered pages so there is no flash of the wrong language.
- Full RTL layout when Arabic is active: set `dir="rtl"` on the document, and rely on CSS logical properties (`margin-inline-start`, `padding-inline-end`, `inset-inline-*`) rather than left/right so the whole UI mirrors. Tailwind v4 supports these directly, plus `rtl:` variants for the cases that need explicit handling.
- Mirror directional UI: sidebar and panels flip sides, chevrons and back/forward arrows flip, dropdown and popover alignment flips, drag handles and resize affordances follow. Do **not** mirror things that shouldn't flip — timeline/Gantt bars still run in chronological order, charts keep their axis conventions, code blocks, file paths, URLs, and API keys stay LTR.
- Handle bidirectional text properly in user content: issue titles, descriptions, comments, and labels get `dir="auto"` so each piece of text takes direction from its own first strong character. This is what stops mixed Arabic/English strings from rendering wrong, and it applies regardless of the user's chosen UI language.
- Localise dates, times, and numbers through `Intl`, including the week starting on Saturday or Sunday depending on region, and correct relative-time strings ("منذ ٣ أيام").
- Handle Arabic pluralisation properly — it has six plural forms, not two, so counters like "3 issues" need an ICU-style plural rule rather than a naive `n === 1` check.
- Verify the pieces that commonly break under RTL: kanban drag-and-drop and column order, the timeline/calendar views, rich-text editing in descriptions and comments, keyboard navigation, and email notification templates.
- Keep the rest of the pipeline language-agnostic: search should match Arabic text, and Arabic input must survive the API, database, and webhook payloads unchanged.

_No response_

### Contribution

ahmedosamaft:
## What

Arabic is now one of the interface languages, and the layout mirrors for it.

`localeDirection` reads the writing direction of the selected language.
`layout.tsx` puts it on `<html dir>` during the server render, so the first
paint is already mirrored and nothing flips after hydration. `providers.tsx`
hands the same value to Radix, which is what gives a menu its arrow-key order
and alignment — the CSS alone does not. Components position themselves with
logical properties (`ms`/`me`, `ps`/`pe`, `start`/`end`, `border-s`/`border-e`,
`text-start`/`text-end`) in place of the physical ones, across about 340 sites.

Three kinds of thing keep their direction on purpose:

- **A pair whose other half has no logical form.** `left-1/2` with
  `translate-x-[-50%]` centres an element; moving one half alone pushes it off
  screen. The switch thumb is the same shape the other way, and travels towards
  the end of its track through an explicit mirrored transform.
- **An edge the caller names.** A sheet asked for on the right is a physical
  choice, and its slide-in animation is physical too.
- **Content that is not prose.** Timelines, charts, code blocks, the JSON
  viewer and the dashboard grid sit in a `dir="ltr"` container: time runs one
  way in every language, Recharts draws to absolute coordinates, code is never
  right to left, and `react-grid-layout` has no right-to-left mode.

Text a user wrote gets `dir="auto"`, so an Arabic comment in an English session
— and the reverse — is laid out by its own script. Chevrons and arrows that
point along the reading direction are mirrored; ones with a fixed meaning are not.

The catalogue is 30 namespaces and 2,316 keys, with all six Arabic plural
categories where English has two. Explanatory strings — help, hints,
descriptions, subtitles — are written in light Egyptian Arabic; buttons, field
names, column headers and status names stay Modern Standard, since those are the
terms a reader matches against the API and the docs. Concept labels carry the
English in brackets: التكاملات (integrations), الوكلاء (agents),
مفاتيح واجهة البرمجة (API keys).

The last hardcoded strings in the activity feed and the schedule presets move
into the catalogue too. A preset's value stays English — it is what the cron
parser reads back, and only its label is shown.

## Why

The interface shipped in four languages, all written left to right, and nothing
in the app read a language's direction. An Arabic-speaking team had no catalogue
to select, and selecting one would have rendered Arabic text in a layout built
the wrong way round.

The string-extraction half of the issue was already done when I picked it up —
only five files still held literals. The work was the other two halves.

Closes #126

## How to test

1. `bun run dev`, then switch the language to العربية (or send
   `Accept-Language: ar-SA` with no cookie).
2. The shell mirrors: sidebar on the right, text right-aligned, menus opening
   from the correct edge. **Reload — there is no left-to-right flash**, because
   `dir` is in the server-rendered HTML. `document.documentElement.dir === 'rtl'`.
3. Open a dropdown and a select: check marks, submenu chevrons and inset padding
   sit on the correct side, and arrow keys move the highlight the right way.
   That last part is what `Direction.Provider` buys.
4. Settings → General: the switches slide towards the end of their track, not
   out of it.
5. The gantt still runs left to right; drag a bar and both ends resize correctly.
   Same for the cycles timeline and the issue status timeline.
6. Charts, code blocks, the JSON viewer and the `⌘K` hints stay left to right.
7. Kanban drag-and-drop: move a card between columns and reorder within one.
8. Type an Arabic comment in an English session and an English one in Arabic:
   each block reads correctly.
9. Switch back to English — everything returns to left-to-right, nothing sticks.

> If you add a locale directory while `next dev` is already running, restart it.
> Translations resolve through a dynamic import, so Next will not see a directory
> that did not exist at boot, and it falls back to English silently.

## Checklist

- [x] `bun run lint` and `bun run format:check` pass — including the new
      `i18n-json` rule, on all five locales
- [x] Tests added: `localeDirection`, and Arabic Accept-Language matching on both
      web and API. Both were mutation-tested — I broke the direction logic and the
      locale list and confirmed each test goes red.
- [x] Docs updated: `docs/dev/i18n.md` gains a Direction section, and
      `apps/web/AGENTS.md` the rule for writing new components
- [ ] Database schema — unchanged. `user_preference.locale` has no CHECK, so a
      new language needs no migration.
- [ ] New environment variables — none

## Screenshots
image


croffasia:
@ahmedosamaft  Thanks for the work here

Before we go further, I'd like to suggest a much simpler route. shadcn supports RTL out of the box:

- https://ui.shadcn.com/docs/components/base/direction
- https://ui.shadcn.com/docs/components/base/sidebar
- https://ui.shadcn.com/docs/components/base/calendar

I tried it on `main`: `dir="rtl"` on `<html>`, `DirectionProvider` in `providers.tsx`, and `side="right"` on the sidebar. Nothing else. The whole app mirrored correctly - board, header, cards, menus, dialogs.

That's three lines instead of rewriting classes across 245 files. From there, let's walk the pages in RTL and fix only what actually looks wrong. What I hit so far: dialogs (close button, header alignment), the filter popover in work items, the floating chat window.

ahmedosamaft:
> @ahmedosamaft Thanks for the work here
> 
> Before we go further, I'd like to suggest a much simpler route. shadcn supports RTL out of the box:
> 
> * https://ui.shadcn.com/docs/components/base/direction
> * https://ui.shadcn.com/docs/components/base/sidebar
> * https://ui.shadcn.com/docs/components/base/calendar
> 
> I tried it on `main`: `dir="rtl"` on `<html>`, `DirectionProvider` in `providers.tsx`, and `side="right"` on the sidebar. Nothing else. The whole app mirrored correctly - board, header, cards, menus, dialogs.
> 
> That's three lines instead of rewriting classes across 245 files. From there, let's walk the pages in RTL and fix only what actually looks wrong. What I hit so far: dialogs (close button, header alignment), the filter popover in work items, the floating chat window.

@croffasia  You're right, and I've restructured around it. The PR is now two commits, 82 files instead of 245.

I did check one thing before dropping the sweep — I wanted to know what the three lines actually cover. Applied them to clean `main` and compiled the CSS: Tailwind's physical utilities compile to physical properties, which don't respond to `dir`.

```
.text-left  ->  text-align: left
.ml-2       ->  margin-left: …
.left-0     ->  left: 0px
.border-l   ->  border-left-width: 1px
```

135 such rules in the bundle. Radix mirrors its own primitives, which is why the first pass reads as "everything moved" — what's left is the padding and alignment inside our own components. The three you found are that same category, not a separate list.

So the foundation is right and the sweep was the wrong way to finish it. Walking pages finds the ones that actually matter; the rest can stay.

**The RTL foundation.** Your three lines, plus the parts a page walk can't surface: the timelines position bars in pixels from the left edge, Recharts draws absolute SVG coordinates, code/JSON/shell blocks are never RTL, and `react-grid-layout` and xyflow both pin themselves LTR for their own coordinate math. Each keeps its direction through a container, and the parts of them that are prose (a dashboard widget, a sticky note) take the page's back.

Two physical properties have to move rather than be replaced: `Switch`'s thumb slides on `translate-x`, which has no logical form, so it travels out of its track instead of across it. The sidebar rail is the same shape.

The three you listed are fixed, plus what I hit walking further: the notes board overlays, and text a user types — `unicode-bidi: plaintext` per block in the editors so a paragraph of Arabic and one of English can sit in the same description, and `dir="auto"` on single-line fields only once there's something to read (empty, `auto` resolves to LTR and strands the placeholder).

**The Arabic locale.** Catalogue, `RTL_LOCALES`, API validator, date-fns. Independent of the layout work if you'd rather review them separately.

croffasia:
@ahmedosamaft  Looks good overall. One thing: editing files in `apps/web/src/components/ui/` is a bad practice - those are native shadcn components and can be overwritten on update. Please check their docs for a way to apply these changes from the outside.

ahmedosamaft:
> @ahmedosamaft Looks good overall. One thing: editing files in `apps/web/src/components/ui/` is a bad practice - those are native shadcn components and can be overwritten on update. Please check their docs for a way to apply these changes from the outside.

Good catch — fixed, and you were more right than I realised: the codebase
already does this. `globals.css` overrides the shadcn overlay shadows through
`data-slot` selectors, with a comment saying the generated components are kept
as-is. I'd broken an established convention rather than an abstract rule.

`src/components/ui/` is now byte-identical to upstream. The five RTL fixes moved
out:

| Was | Now |
| --- | --- |
| `dialog`, `sheet` close buttons | `globals.css` → `[dir='rtl'] [data-slot='dialog-close']` |
| `switch` thumb travel | `globals.css` → `[data-slot='switch-thumb'][data-state='checked']` |
| `sidebar` inset margin + trigger icon | `globals.css` → `[data-slot='sidebar-inset']` / `sidebar-trigger` |
| `sidebar` `side` default | `useSidebarSide()`, passed by the two callers |
| `chart` `dir="ltr"` | the two `ChartContainer` call sites |

One thing worth recording, because it isn't obvious and it bit me: specificity is
not the lever. Tailwind v4 emits `@layer theme, base, components, utilities`, so a
rule in `@layer components` loses to a utility class *no matter how specific it
is*. These work because they're unlayered, which outranks every layer. Verified in
the compiled output — the switch utility lands at line 4655 inside
`@layer utilities`, the override at 6288 unlayered.

Two of them couldn't be pure CSS, so they went through the public API instead:

- **`Sidebar`** already takes `side`, so the two callers pass it. Cleaner than CSS
  and it drops a `radix-ui` import from the primitive.
- **`chart`** needs `dir="ltr"` as an *attribute* — Tailwind's `rtl:` variant keys
  off `[dir="rtl"]`, so pinning a subtree needs the real attribute and CSS can't
  set one. `ChartContainer` spreads `...props`, so the two call sites pass it.

Documented the rule in `apps/web/AGENTS.md` and the mechanism in `docs/dev/i18n.md`
so this doesn't regress.

One thing I'd flag:

1. **`dialog`'s `sm:text-left`** is the one I'd push back on slightly. It's not an
   RTL override, it's a wrong default — a dialog header should be `text-start` in
   any language. Overriding it externally means restating the `sm:` breakpoint in
   CSS, which is uglier than the one-word fix. I've done it externally for
   consistency, but it may be worth an upstream issue.

PR #160, opened 2026-08-15
fix(web): generate uuids without crypto.randomUUID over plain http

h-mbl:
## What

Adds a shared `uuid()` helper in the web app and routes note stickers, agent chat messages
and their queue, chat session tabs, filter conditions, custom field options, and dashboard
widget creation through it. It uses `crypto.randomUUID()` when available and otherwise
builds an RFC 4122 v4 id from `crypto.getRandomValues()`.

## Why

`crypto.randomUUID()` exists only in a secure context (HTTPS or localhost). On a
self-hosted instance served over plain HTTP on a LAN it is `undefined`, so creating a note,
opening the chat panel, sending an agent chat message, adding a filter, a custom field
option, or a dashboard widget crashes with `crypto.randomUUID is not a function`.
`crypto.getRandomValues()` is a CSPRNG available in every context, so the ids keep the same
randomness quality.

## How to test

Serve the app over plain HTTP on a non-localhost origin (e.g. a LAN IP). Before: creating a
note throws `crypto.randomUUID is not a function` in the console, and opening the chat panel
(hotkey `c`) breaks the whole page. After: notes, agent chat, filters, custom field options
and dashboard widgets create normally.

## Checklist

- [x] Tests added or updated for the changed behaviour — `apps/web/src/utils/uuid.test.ts` covers the normal path, the fallback with `randomUUID` undefined, and uniqueness
- [ ] Database schema changed — n/a, no schema change
- [ ] New environment variables documented in `.env.example` — n/a, no new variable
- [ ] Docs updated — n/a, no user-facing docs affected

### DIFF
--- a/apps/web/src/features/ai-chat/hooks/useChatSessions.ts
+++ b/apps/web/src/features/ai-chat/hooks/useChatSessions.ts
@@ -2,6 +2,7 @@
 
 import { useCallback, useEffect, useState } from 'react';
 import { arrayMove } from '@dnd-kit/sortable';
+import { uuid } from '@/utils/uuid';
 
 // One open conversation in the chat panel. `threadId` is null until the agent answers
 // the first message, and stays null for an agent that keeps no memory. `running` and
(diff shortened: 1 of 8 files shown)


henry2man:
@h-mbl take a look here ;) https://github.com/croffasia/itsaplan/blob/main/CONTRIBUTING.md#sign-the-contributor-license-agreement

@croffasia This also affects to chat feature (hotkey `c`) in non https hosted environments (whole page breaks if you try to open the chat)...

h-mbl:

croffasia:
Nice catch, thanks! Merging 🙏

PR #227, opened 2026-08-26
feat(god): instance setting mcpEnabled as enabled by default

### LINKED ISSUE #224: [feature]: instance setting to enable MCP on new projects by default
### Problem

MCP access is a per-project toggle that defaults to off. On an instance where the whole point is driving projects through MCP, every new project starts invisible to the connected client: it won't appear in list_projects until someone remembers to open project settings and flip the toggle.

That's a sensible default for a team instance where MCP is the exception. It's friction on an instance where MCP is the rule, and forgetting the toggle looks like a broken integration rather than a setting.

### Proposed behaviour

An instance-wide setting in god mode: "Enable MCP on new projects." Off by default, so nothing changes for existing installs.

When it's on, newly created projects start with MCP enabled. Existing projects are untouched either way, and the per-project toggle keeps working exactly as it does now — this only changes what a fresh project starts with.

Flipping the toggle manually per project, which is what I do now across 8 projects. It works, it's just a step that's easy to forget and only discoverable by noticing a project missing from the MCP client.

Could also be a per-project-creation option (a checkbox in the new project dialog), but an instance default matches how the operator actually thinks about it: on a given install, MCP is either how you work or it isn't.

### Contribution

ryakel:
## What

Adds a General section to god mode holding the instance-wide project defaults, with the first one being whether a new project is reachable over MCP.

## Why

On an instance driven through MCP, every new project starts invisible to the connected client until someone remembers the per-project toggle, which reads as a broken integration rather than a setting.

Design follows @croffasia's direction in the issue: its own section under General, default on.

The defaults live in `app_setting` under the `projects` key, following the storage settings pattern, and `createProject` reads them on the insert. Projects that already exist are untouched, and the per-project toggle is unchanged.

`mcpEnabled` gates visibility on top of membership rather than granting access, so a project still only appears to a member holding a key.

Closes #224

## How to test

```bash
cp .env.example .env   # generate APP_ENCRYPTION_KEY, BETTER_AUTH_SECRET, WORKER_INTERNAL_TOKEN
docker compose build api web && docker compose up -d
```

1. Register the first account (becomes the instance owner) and open **God mode → General**.
2. Create a project. Its **Project settings → MCP Server** shows MCP access already enabled.
3. Turn the toggle off in General, create a second project. That one comes up with MCP access off.
4. The first project is still on, since a change here reaches new projects only.

API directly:

```
GET  /god/project-defaults        -> {"mcpEnabled": true}
PUT  /god/project-defaults        {"mcpEnabled": false}
GET  /projects/:key/settings      -> reflects the default at creation time
```

## Upgrade behaviour

Worth a deliberate call, since this changes what existing installs do.

The setting is absent until someone saves it, and an absent key reads as the default. So after upgrading, an instance that never opens God mode gets MCP-on for projects created from then on, not just fresh installs. Projects that already exist keep whatever they're set to.

That is a visibility default rather than an access grant, so a project still only appears to a member holding a key. It is still a silent change on upgrade, so flagging it rather than deciding it for you.

If you'd rather upgrades keep today's behaviour, a migration can seed `{"mcpEnabled": false}` only when the instance already has projects, which distinguishes an upgrade from a fresh install. Storage settings set the precedent for seeding (migration 0046). Happy to add it if you want it.

## Checklist

- [ ] Database schema changed: migration generated with `bun run db:generate` and committed. n/a, no schema change; this only changes what `mcpEnabled` is set to on insert
- [ ] New environment variables documented in `.env.example`. n/a, no new env vars
- [ ] Docs updated (`README.md` or the relevant `AGENTS.md`). n/a, the settings and god-section patterns are documented in code comments rather than AGENTS.md

Six existing tests encoded MCP-off as the default. The ones asserting the default now assert the new one; the MCP-blocking tests set `mcpEnabled: false` explicitly, since they need a disabled project rather than the default specifically.

New coverage in `god-project-defaults.test.ts`: access control, the read/write round trip, the default landing on new projects, existing projects staying untouched, and a non-owner's project picking it up.

Strings added to all five locales in English. Happy to drop the four non-English ones if you'd rather translate them yourself.

## Screenshots
Screenshot 2026-08-25 at 7 28 31 PM
Screenshot 2026-08-25 at 7 28 59 PM

### DIFF
--- a/apps/api/src/modules/god/__tests__/integration/god-project-defaults.test.ts
+++ b/apps/api/src/modules/god/__tests__/integration/god-project-defaults.test.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect, beforeEach } from 'bun:test';
+import { resetDb } from '#tests/helpers/db';
+import { addUser, setup } from '../helpers';
+
+// The instance-wide project defaults under god mode: what a project starts with
+// when it is created. Only the instance owner may read or change them, and a
+// change reaches new projects only — the ones that already exist keep whatever
+// they were set to.
+
+describe('god project defaults', () => {
+  beforeEach(async () => {
+    await resetDb();
+  });
+
+  describe('access', () => {
+    it('refuses both routes for a user who is not the instance owner', async () => {
+      await setup();
+      const outsider = await addUser();
+
+      expect((await outsider.api.god['project-defaults'].get()).status).toBe(403);
+      expect((await outsider.api.god['project-defaults'].put({ mcpEnabled: false })).status).toBe(
+        403,
+      );
+    });
+  });
+
+  describe('settings — GET/PUT /god/project-defaults', () => {
+    it('starts with MCP on, so an instance driven through MCP needs no toggle', async () => {
+      const { god } = await setup();
+
+      const res = await god.api.god['project-defaults'].get();
+
+      expect(res.status).toBe(200);
+      expect(res.data?.mcpEnabled).toBe(true);
+    });
+
+    it('stores a change and reads it back', async () => {
+      const { god } = await setup();
+
+      const put = await god.api.god['project-defaults'].put({ mcpEnabled: false });
+      const get = await god.api.god['project-defaults'].get();
+
+      expect(put.status).toBe(200);
+      expect(put.data?.mcpEnabled).toBe(false);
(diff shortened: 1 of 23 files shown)


ryakel:

croffasia (approved):

PR #228, opened 2026-08-26
feat: import issues from files via agent chat

### LINKED ISSUE #114: [feature]: let AI agents understand uploaded files and import spreadsheet tasks
### Problem

AI agents cannot read or understand uploaded files such as Excel and Word documents. An Excel workbook may already contain a task list, but users currently have to  
  recreate every task manually in It's a Plan.

### Proposed behaviour

Allow internal AI agents to read supported attachments such as Excel and Word files. When a spreadsheet contains a task list, the agent should extract the rows, map 
  the columns to It's a Plan fields, show a preview, and create the tasks after user confirmation.

_No response_

### Contribution

netoxp70:
## What

Lets an internal agent turn an uploaded spreadsheet or document into issues, with explicit human confirmation before anything is created (Closes #114):

- **Upload**: the chat composer gains a paperclip (.xlsx, .csv, .docx). The file goes to the object store under `projects/<id>/imports/` and creates a draft row in the new `issue_import` table.
- **Agent tools** (opt-in capabilities, like every other agent tool): `read_import_file` returns headers plus sample rows; `save_import_mapping` stores which column feeds which issue field (title, description, priority, dueDate, labels, assignee).
- **Review card**: when the mapping is saved, the agent ends its reply with a ```issue-import fence and the chat renders a card — mapped columns with field badges, real preview rows, and Confirm / Cancel buttons. Same mechanism as ```chart.
- **Confirm**: a REST route re-reads the stored file, applies the saved mapping server-side, and creates the issues through the same service the UI uses (sequence numbers, WIP limits, validation). Rows that do not resolve are reported as skipped.

Deliberate choices:

- **No `assistant-ui`**: the existing AG-UI chat already renders rich blocks via markdown fences (`create_chart`); this follows that pattern instead of introducing a second chat stack.
- **The model cannot create issues by itself**: creation lives behind the confirm route only. It also means rows come from the file, never from a model transcription of it.
- **No legacy `.xls`**: ExcelJS reads xlsx/csv; SheetJS on npm has known advisories and its fix is distributed off-npm, which would break the lockfile/Docker install. An .xls file gets a "save as .xlsx" hint.

## Why

Closes #114. A spreadsheet that already holds a task list should not have to be retyped by hand.

## How to test

- Unit: `bun test apps/api/src/modules/imports/__tests__/unit/parse-mapping.test.ts` (csv quoting/delimiters, mapping validation, label/assignee/date resolution, skipped rows).
- Integration: `apps/api/src/modules/imports/__tests__/integration/imports.test.ts` covers upload validation, member-only access, map → confirm → created keys, double-confirm 409, cancel flow. Needs MinIO like the attachments suite.
- End to end: enable the two import tools on an agent, attach a spreadsheet in the agent chat, ask it to import the tasks; review the card, confirm, see the issues on the board.

## Checklist

- [x] `bun run typecheck`, `bun run lint` and `bun run format:check` pass
- [x] Database schema changed - migration `0106_violet_manta.sql` committed
- [ ] New environment variables documented in `.env.example` - n/a, no new variable
- [x] Docs updated (tool catalog copy)


croffasia:
@netoxp70  

Thanks, the flow itself is right. Two things I would like changed before this lands.

1. Let's add a table for uploads, something like `chat_attachments`, holding just the file: `public_id`, `project_id`, `uploaded_by`, `s3_key`, `filename`, `content_type`, `size_bytes`, no status. Right now `issue_import` is the file and the import state at once, so a file cannot exist unless it is an import. With a separate table we can use uploads for more than issue imports later, for example reading a spec or a log dropped in the chat.

Expose upload and read as normal API routes tagged with `mcpTool()`, and put their keys in `ALWAYS_ON_ACTIONS` so every internal agent has them and MCP clients get them too. Access stays inside the project and follows what attachments do: `work_items:read` to read, `work_items:create` to upload.

2. Let's keep one agent tool for the import itself. Reading the file is covered by the read tool above, so the import needs only the mapping step: one tool that takes the attachment id and the column mapping and produces the draft the review card renders. Issues are still created by `confirm` alone, when the person presses the button, so that route stays off the tool list for internal and external agents both.

3. One more thing, and it would be good to have it here. The marker we put in the message, `[file for import: "Jira.csv" (import id: 7e01be81-...)]`, is shown to the user as raw text with a uuid in it. Let's render it the way charts and the import card are already rendered in `lib/markdown.ts`: replace the marker with the file name and a download link. That needs a download route for the attachment, by `public_id`, like issue attachments have.

netoxp70:
@croffasia

All three points are done in the latest push.

**1. Separate `chat_attachment` table + upload/read routes**

Added `chat_attachment` (`public_id`, `project_id`, `uploaded_by`, `s3_key`, `filename`, `content_type`, `size_bytes`) — no status, just the file. `issue_import` now references a `chat_attachment` by its `public_id` instead of storing the file itself.

Upload and read are exposed as normal API routes tagged with `mcpTool()`, and their keys are in `ALWAYS_ON_ACTIONS` so every internal agent and MCP client gets them. Access follows the existing attachment permissions: `work_items:read` to read, `work_items:create` to upload. A raw download route by `public_id` is public (same as issue attachments) so the file marker can link to it without a session.

Project storage quota now counts chat attachment bytes alongside issue attachment bytes, so uploads are bounded by the same limit.

**2. One import tool for the mapping step**

`read_import_file` is gone — the generic read tool covers it. The import keeps a single tool that takes the attachment id and the column mapping and produces the draft the review card renders. Issue creation still lives behind the `confirm` route only; it is not on the tool list for internal or external agents.

**3. File marker rendered as a download link**

The marker `[file: "Jira.csv" (attachment id: 7e01be81-...)]` is now rendered as the file name with a download link, both in agent messages (markdown) and in user messages (plain text). The regex and URL builder are exported from `lib/markdown.ts` and shared by both renderers, so the link is consistent across the two code paths.

Tests: unit tests cover CSV quoting/delimiters, mapping validation, and skipped rows; integration tests cover upload validation, member-only access, read → map → confirm → created keys, double-confirm 409, and the cancel flow. The full suite passes in the Docker test gate and in CI.

croffasia:
Thanks for this, @netoxp70 . Great work. The separate attachments table gives  us a base to build on - we'll keep expanding what the chat can do with files.

PR #229, opened 2026-08-26
feat(api): expose checklist routes as MCP tools

### LINKED ISSUE #226: [feature]: checklist tools for AI agents
### Problem

Checklists are on the card, but agents cannot touch them. Every other part of an issue an agent works with is reachable over MCP — description, comments, labels, state, time entries, attachments — and checklists are the one gap. There is no tool for them, and an API key only authenticates at `/mcp`, so the REST routes are not an escape hatch either.

What this looks like in practice: an agent was asked to expand a two-item checklist to seven. It could not, so the list went into the issue description as markdown checkboxes instead. That works, but the card already has a real checklist, so the same list now lives in two places and the card's own one is the stale copy. Expanding a checklist is exactly the repetitive step you would hand to an agent.

### Proposed behaviour

An agent can manage the checklist on an issue it already has edit rights to:

- add a checklist to an issue
- add an item to a checklist
- change an item's text, tick it, or untick it
- remove an item

Reading needs nothing new — `get_issue` already returns an issue's checklists with their items.

Rights behave exactly as they do for every other issue write: an agent that cannot edit the issue cannot touch its checklist, and a project with MCP switched off stays closed.

The current workaround is keeping the list in the description as markdown checkboxes. It works, but it splits one list across two places on the same card and leaves the native checklist as the stale copy.

Deliberately not proposed: reordering checklists or items, which is a drag in the UI rather than something an agent needs, and deleting a whole checklist, which cascades to items a person wrote — a wider blast radius than the need justifies. An agent that can remove its own bad item can already undo its own mistake.

Implementation note, in case it helps you judge the size: the routes already exist and carry their own argument schema and permission checks. What is missing is only the `x-mcp` tag that `generateRouteTools` filters on. On my fork this is four one-line tags in `apps/api/src/modules/issues/index.ts`, plus one line in the server instructions saying a `checklistId` comes from `get_issue` rather than `get_project` (the surrounding text says every id comes from `get_project`, which would send an agent looking in the wrong place). Format, lint, typecheck and the full container test suite pass on it. Happy to send it as a pull request, or to change the shape if you would rather it looked different.

### Contribution

iMofas:
## What

Tags four checklist routes as MCP tools, so an agent can add a checklist to an issue, append items, tick them off, and remove an item. No new logic — the argument schema and the permission check come from each route.

Also adds a line to the server instructions saying a `checklistId` comes from `get_issue`. The paragraph right above it says every id comes from `get_project`, so without this an agent looks confidently in the wrong place and the new tools are unusable.

Left untagged on purpose: the checklist list (`get_issue` already returns checklists with their items, so a separate tool is just another fork in the road for every agent on every call), the two reorder routes (that's a drag in the UI), and deleting a whole checklist (it cascades to items a person wrote, which is a wider blast radius than an agent needs — being able to remove its own bad item is enough to undo its own mistake).

## Why

Checklists were the one part of an issue an agent could not touch. Description, comments, labels, state, worklogs and attachments are all reachable over MCP; checklists had no tool at all, and an API key only authenticates at `/mcp`, so the REST routes are not a way in either.

What it looked like in practice: an agent was asked to expand a two-item checklist to seven, couldn't, and put the list in the issue description as markdown checkboxes instead. The card already had a real checklist, so the same list ended up in two places with the native one stale.

Closes #226

## How to test

Point an MCP client at a project that has MCP enabled, authenticating as a project agent.

1. `get_issue` on any issue — the response already carries `checklists`, each with its items.
2. `create_checklist` with that issue id.
3. `create_checklist_item` with the `checklistId` from step 2.
4. `update_checklist_item` with `done: true`, then look at the card in the UI.
5. `delete_checklist_item` on the same item.

For the access side, turn MCP off for the project and repeat step 3 — it 403s, because the `checklist` guard runs `assertMcpAllowed` like every other issue write. Same for a member whose role lacks `work_items` edit.

## Checklist

- [x] Tests added or updated for the changed behaviour — `checklists.test.ts` pins which checklist routes are tagged and which stay session-only, using the existing `untaggedRoutes` helper
- [ ] Database schema changed — n/a, no schema change
- [ ] New environment variables — n/a, none
- [ ] Docs updated — n/a; the wording an agent reads lives in `SERVER_INSTRUCTIONS`, which this change updates

## Screenshots

n/a — no UI change.


iMofas:

croffasia:
@iMofas Let's also go ahead and add this to the actions catalog for internal agents. AGENT_ACTIONS (apps/api/src/modules/agents/core/runtime/tools/catalog.ts) under the issues group

iMofas:
Added, thanks — all four under the issues group, opt-in like the rest.

One thing I checked while doing it: `get_issue` is in ALWAYS_ON_ACTIONS and already returns an issue's checklists with their item ids, so an agent granted just `create_checklist_item` can find the checklist it needs. No companion read to wire up.

Unrelated, take it or leave it: nothing catches a typo in a catalog key. It has to match a route tool name, but the only cross-check against the route table is the `hidden` args assertion in route-tools.test.ts, so a misspelled key would just be a toggle in the UI that grants nothing. Happy to add a test that every AGENT_ACTIONS / ALWAYS_ON_ACTIONS key resolves in routeTools(), here or separately — your call.

croffasia (approved):

PR #250, opened 2026-08-28
feat: add generic OIDC sign-in, SCIM 2.0 provisioning, and a password-auth switch

### LINKED ISSUE #161: [feature]: generic OIDC/OAuth2 support
### Problem

It would be great to have OIDC/OAuth2 support in case one is selfhosting KanIDM, GitLab, Forgejo or other OIDC providers.
`better-auth` is already a dependency and it ships a `genericOAuth` plugin that accepts a `discoveryUrl`.
So no new dependency or handwritten OAuth flow is needed.

### Proposed behaviour

Being able to authenticate via OIDC

_No response_

### Contribution

- [ ] I am willing to open a pull request for this

bacali95:
## What

Adds three related pieces of authentication infrastructure, all built on top of what
`better-auth` already ships in this repo:

- **Generic OIDC/OAuth2 sign-in** — one configurable provider, discovered from its
  `.well-known/openid-configuration` document. Works with Keycloak, Authentik, KanIDM,
  GitLab, Forgejo, Okta, Entra, or any other OIDC-compliant identity provider.
- **A switch to turn off email/password sign-in** — usable once a working SSO provider
  (Google or the new OIDC one) is configured, so an instance can run entirely off its
  identity provider. The API refuses to disable it while no provider is usable, so an
  instance can never be left with no way in.
- **SCIM 2.0 provisioning** (`/scim/v2`, RFC 7643/7644) — an identity provider can
  create, update, and deactivate accounts, and its groups can be mapped to project
  membership in god mode. A group can also reach the instance two other ways without a
  separate `/Groups` push: embedded on a SCIM user's `groups` attribute, or as the
  `groups` claim on an OIDC sign-in's ID token — both funnel into the same group/mapping
  mechanism.

## Why

Closes #161. Self-hosting an instance behind an existing identity provider is a common
ask for anything aimed at organizations, and `better-auth`'s `genericOAuth` plugin
already covers the OIDC side without a new dependency. SCIM and the password-auth switch
extend the same idea to full account lifecycle management (provisioning, deprovisioning,
group-based access) and to instances that want the identity provider to be the *only*
way in, not just an alternative to the password form.

### Design notes

- **Credentials live in the database, not env** (`app_secret`, AES-256-GCM via
  `@repo/crypto`), matching the existing Google-provider pattern exactly — the operator
  changes them from god mode without a restart. No new environment variables.
- **One OIDC provider per instance.** `better-auth`'s generic-oauth config array is
  materialized once at startup, so the provider id is a fixed constant (`"oidc"`) and its
  config object is mutated in place, refreshed per request — the same by-reference trick
  the existing Google provider already uses.
- **SCIM is mounted on the root app, not the session-gated planner.** It authenticates
  with its own bearer token and answers its own SCIM error document
  (`urn:ietf:params:scim:api:messages:2.0:Error`) rather than the planner's `{ error }`
  shape.
- **Provisioning claims an existing account by email** (case-insensitive, since
  better-auth does not normalize case at sign-up) instead of creating a duplicate when
  the person already has an account — from a prior invite, a password sign-up, or an
  earlier OIDC sign-in.
- **Group membership has one clear owner per row.** `project_member.source` distinguishes
  a membership that came from an invite from one a SCIM/OIDC group sync created; the sync
  only ever writes, re-roles, or removes rows it owns, so a membership someone set up by
  hand is never touched by it, and the reverse — editing a synced membership by hand — is
  refused (409) because the next sync would silently undo it.
- **Deactivation covers every credential**, not just the session: an inactive account
  gets a 401 from `authContext` (every planner route, MCP included), and
  `databaseHooks.session.create.before` refuses to open a new session for it regardless
  of sign-in method.

## How to test

Local, with a throwaway Keycloak or Authentik in Docker (or KanIDM):

1. `bun run db:migrate` to pick up the new tables/columns
   (`scim_group`, `scim_group_member`, `scim_group_mapping`, `user.active`,
   `user.scim_external_id`, `project_member.source`).
2. God mode → Integrations → Auth provider: paste the discovery URL, client ID and
   secret, copy the shown redirect URI into the identity provider's client, name the
   button, and enable it.
3. Sign out. The sign-in screen now shows the SSO button. Sign in through it — a new
   account is created (subject to the registration mode under Authentication) and lands
   on `/`.
4. God mode → Authentication: turn off "Email and password". Confirm the switch explains
   itself and stays disabled until a working SSO provider exists. With it off, the
   sign-in screen shows only the SSO button, and `POST /api/auth/sign-in/email` answers
   403.
5. God mode → Integrations → SCIM: enable, generate a token (shown once), copy the SCIM
   endpoint it displays.
6. `curl -H "Authorization: Bearer <token>" <API_URL>/scim/v2/ServiceProviderConfig`
   should return the discovery document; `POST /scim/v2/Users` should provision an
   account; `PATCH` it with `{"Operations":[{"op":"replace","path":"active","value":false}]}`
   should end its session.
7. Point a real identity provider's SCIM app at `<API_URL>/scim/v2` with the token and
   confirm users and groups sync through. Open a synced group in god mode, map it to a
   project and a role, and confirm its members appear on the project's Members page with
   a "Managed by SCIM" badge — and that removing someone from the group at the provider
   removes them from the project.
8. For the embedded-groups path: configure the identity provider to add a `groups`
   claim to its ID token (or a `groups` attribute on the SCIM user payload) and confirm
   the same group/mapping mechanism picks it up without ever calling `POST /Groups`.

Automated: `apps/api` gets 77 new integration/unit tests under
`apps/api/src/modules/scim/__tests__/` and `apps/api/src/mcp/__tests__/`, covering the
SCIM discovery endpoints, user and group CRUD, PATCH semantics, the reconciliation rules
above (including the last-owner guard and the invite-vs-SCIM ownership boundary), the
OIDC/password-toggle settings API, and account deactivation across `authContext`, `/me`,
and the MCP handshake.

## Checklist

- [x] New environment variables documented in `.env.example` — none added; every new
      credential (OIDC client secret, SCIM token) is stored encrypted in the database,
      the same way the existing Google credentials are.
- [x] Docs updated (`README.md` or the relevant `AGENTS.md`) — `packages/auth/AGENTS.md`,
      `apps/api/AGENTS.md`, and `docs/self-hosting.md`.

## Screenshots

- OIDC configuration page
Screenshot 2026-08-28 at 18 20 15

- SCIM 2.0 configuration page
Screenshot 2026-08-28 at 18 20 26

- SCIM 2.0 groups mapping form
Screenshot 2026-08-28 at 18 34 07

- New disabled username/password authentication method toggle
Screenshot 2026-08-28 at 18 42 03

- Login page with disabled username/password authentication method
Screenshot 2026-08-28 at 18 25 09

- List of users logged in with OIDC auth provider and imported via SCIM 2.0.
Screenshot 2026-08-28 at 18 24 33


bacali95:

croffasia:
Great work on this @bacali95 thanks! A couple of comments after testing it against a live SCIM run.

1. SCIM can change the account of the instance owner, and the owner cannot undo it. Also deactivating the owner (`user.active = false`) kills god mode, and no route sets it back - only SQL.

2. Removing someone from a group in Okta does not take their access away. The admin removes the user in the identity provider, Okta reports success, and the user keeps the project access the group grants. Nobody is told. Okta sends the removal in the form the SCIM standard defines (RFC 7644 §3.5.2.2) and the API ignores it;

3. Provisioning the same person twice wipes what the identity provider knows about them. A repeated create answers "created" instead of "already exists" (409, per RFC 7644 §3.3) and clears the fields the second request left out, including the id that links the account to the provider. Okta retries a create after a timeout, so this happens on its own.

bacali95:
Thanks for testing this against a live run, @croffasia — all three are fixed, each as its own commit:

1. **Instance owner protection** (`427f9a2`): `createScimUser`/`updateScimUser` now refuse a `god`-role account outright (409). SCIM can no longer rename, reassign, or deactivate the instance owner.
2. **Group member removal via path filter** (`a6b9f13`): the PATCH handler now reads RFC 7644 §3.5.2.2's `path: 'members[value eq "<id>"]'` form — the one Okta sends — instead of only the `value`-carrying form. Removing someone from a group now takes their project access away.
3. **Duplicate create no longer overwrites the link** (`e0dceed`): a repeated `POST /Users` for an address already linked to the provider (`scimExternalId` set) now answers 409 `uniqueness` instead of silently overwriting the link — including the id — with whatever the retry carries. An address that predates the sync still gets claimed as before.

Each fix has its own test (red before the fix, green after), and the full SCIM/god/members/MCP suite (174 tests) plus `format:check`/`lint`/`typecheck` pass. Pushed to `feat/oidc-scim`.

croffasia (approved):

croffasia:
@bacali95 All three check out - merging. Thanks for the fast turnaround!

PR #260, opened 2026-08-30
feat: add email provider test

### LINKED ISSUE #257: [feature]: An email provider health check or test button
### Problem

Testing the mail provider is not straightforward, had to enable sign-in links and use them to debug the process.

### Proposed behaviour

A single button at the bottom of the provider config card, grayed out when the provider is disabled or has no credentials provided. It sends a test email to the signed-in user's own address; no input field needed.

There's probably smarter UX than what I proposed but I'm far from a good UX designer.

### Contribution

grootbro:
Closes #257

## What changed

- adds an owner-only endpoint that sends a real message through the current instance email form
- tests edited SMTP or Resend values before they are saved
- reuses an existing encrypted secret when the secret field is left blank, without persisting the draft
- sends only to the signed-in instance owner; the recipient is not user-controlled
- disables the UI only while the current form is incomplete or a test is already running
- maps provider failures to actionable, sanitized errors while keeping raw details in server logs
- applies a 15-second default SMTP test timeout, including socket inactivity
- adds API coverage for authorization, missing configuration, unsaved drafts, and stored-secret reuse

## Checks

- `bun test apps/api/src/modules/god/__tests__/unit/email-test.test.ts`
- API, Web, and auth package typecheck
- API, Web, and auth package lint
- Prettier and `git diff --check`

The database-backed integration cases are included in the branch and run in GitHub CI.


croffasia:
Works great - let's just stick to the common practice here: let the user send a test email before saving the new values :)

grootbro:
Addressed the unsaved-settings flow in e74d694.

The test action now sends the current form values directly to the owner-only endpoint, merges omitted secret fields with the encrypted saved value in memory, and does not persist the draft. The button is available as soon as the current form is complete, so SMTP and Resend can both be verified before Save.

I also added regressions for a draft SMTP test not changing storage and for reusing a saved secret while previewing edited settings. API/Web/auth typecheck and lint pass; the focused error-classification tests pass. The database-backed integration cases are ready for CI.

croffasia (approved):

croffasia:
Perfect! Merged

PR #270, opened 2026-08-31
feat(web): add French translation

### LINKED ISSUE #255: [feature]: Add French translation
### Problem

It would be great to add French (fr) as a supported language.

The application already supports multiple languages, so this would mainly involve adding and maintaining the French translations for the existing interface.

I'm a native French speaker and I'd be happy to work on the translation and open a PR for it.

### Proposed behaviour

Add French (fr) to the list of supported languages
Translate the existing UI strings into French
Add French to the language selector
Make sure the French locale follows the existing localization structure

_No response_

### Contribution

Rasaboun:
## What

Adds French (`fr`) as a supported interface language. All 30 namespaces under `messages/fr` are translated — 2641 keys, no gaps.

Follows the four steps in `docs/dev/i18n.md`:

- `messages/fr/*.json`
- `fr` in `LOCALES`, with `Français` and 🇫🇷 (not added to `RTL_LOCALES`)
- the `fr` locale from `date-fns`, so the calendar's month and weekday names follow
- `fr` in the API's `LOCALES` and the `Locale` union, so the preference can actually be saved

Two small things beyond the four steps:

- `accept-language.test.ts` and the API's `locale.test.ts` both used `'de-DE,fr;q=0.9'` as their example of a request for a language the instance *doesn't* support. That assertion stops being true once `fr` ships, so they ask for `ja` now.
- The language list in `README.md`.

## Why

Closes #255.

## Translation notes

A few decisions worth a second opinion from another French speaker, since they run across all 2641 keys:

- **"issue" → « ticket »**, and **"work item" uses the same word** rather than a second one (`nav.workItems` and `permissions.resources.work_items` are both "Tickets"). English carries two names for one thing here; French normally just has the one. Happy to split them if you'd rather keep the distinction visible.
- **No narrow/non-breaking space before `? ! : ;`** — a plain space, applied uniformly. Strict French typography wants U+202F, but it's invisible in diffs and the next person to edit a string will type a normal space, so the file ends up mixed. Consistency seemed worth more here.
- Other terms held steady throughout: état / statut, étiquette, responsable / délégué, compétence (skill), checklist, backlog, story points, token, webhook, tableau (board) vs tableau de bord (dashboard). "God mode" stays in English, matching `roleGod: "God"`.
- Where a placeholder holds a noun, the phrasing avoids depending on its gender. `featureOffTitle` is "La section {feature} est désactivée", because the seven feature names are a mix of masculine and feminine.

## How to test

1. `bun run dev`, then pick **Français** in the language switcher (or set `NEXT_LOCALE=fr`).
2. Walk a board, an issue, the inbox and Settings — Settings is the biggest namespace and the most likely place to spot an awkward phrase.
3. Open the date picker on an issue to confirm the month and weekday names are French, not English.
4. `Accept-Language: fr` on a signed-out page should render French without a cookie.

Reporting anything that reads badly is genuinely useful — I read every string in context, but a second pair of eyes on the long Settings hints would help.

## Checklist

- [ ] Tests added or updated for the changed behaviour — two existing tests updated; the message files are covered by `eslint-plugin-i18n-json` (`valid-message-syntax`, `identical-keys`) via `bun run lint`
- [ ] Database schema changed — n/a, `user_preference.locale` is plain `text` with no CHECK
- [ ] New environment variables documented — n/a
- [x] Docs updated (`README.md`)

## Screenshots

Not a visual change — same layout, French strings. Shout if you'd like before/after shots of a particular screen.

## Two pre-existing things I noticed and left alone

Both affect the existing languages too, so they felt out of scope here — happy to open separate issues:

- `features.timeLogging` is used at `IssueDetailContent.tsx:184` but isn't in `messages/en/common.json`. Adding it to `fr` alone would fail `identical-keys`, so it needs fixing in `en` first.
- Notification subjects and bodies are hardcoded English in `apps/api/src/modules/notifications/outbound.ts` (lines 81–113), with no locale lookup for any language. So `account.preferences.languageDescription` — "la langue de l'interface, de vos e-mails et du bot Telegram" — currently overpromises for ru/uk/zh/ar as well as fr.


Rasaboun:

croffasia (approved):

croffasia:
@Rasaboun  Nice work, thanks! Merged - it'll go out in the next release.

PR #274, opened 2026-09-01
feat: add project Docs and watcher management

grootbro:
## Summary

- add native project Docs with a hierarchical page tree, rich editor, favorites, private pages, archive/restore, copies, subpages, export, and version history
- link Docs and work items bidirectionally from either surface, with searchable pickers, live refresh, private-page filtering, archived-state handling, and combined permissions
- integrate Docs with project permissions, feature settings, command navigation, agents, project copying, account deletion, SCIM, protected assets, and storage quotas
- let users with work-item edit access add or remove other human project members as issue watchers while preserving self-watch behavior
- keep the editor toolbar visible while scrolling and make favorite documents directly accessible and manageable from the Docs list

## Data and safety

- store document/work-item links as an explicit relation with same-project validation in both the API and PostgreSQL
- use optimistic document versions to prevent silent overwrites
- scope private documents, linked-item visibility, and revision history to their owner
- sanitize portable HTML exports and downgrade active asset MIME types such as SVG to inert downloads
- validate document assets and quota updates server-side
- preserve document ownership and revisions across project copies and account deletion
- validate watcher targets against project membership and permissions

## Checks

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `API_URL=http://localhost:3000 bun run build`
- 170 Web tests, including export-security regressions
- locale key parity across English, Arabic, French, Russian, Ukrainian, and Simplified Chinese
- downstream self-hosted production smoke test for Docs, favorites, sticky toolbar, both link pickers, and bidirectional display

The API integration suite is left to the required GitHub CI because the local Docker daemon/test PostgreSQL service was unavailable.


croffasia:
@grootbro Had a look at the feature - great job, this turned out really well. Let's open it up for broader testing and iterate from there if anything comes up. Thanks for pushing this through!

grootbro:
@croffasia Thanks for the review and merge! Glad you like it 🙌 We're using Docs with our team too, so we'll bring back feedback and fixes as things come up.

PR #288, opened 2026-09-04
fix(web): keep the board scrollable while the issue detail panel is open

### LINKED ISSUE #287: [bug]: the board cannot be scrolled while the issue detail panel is open
### What happened

The issue detail panel renders its wrapper as `fixed inset-0 z-50 flex`, so it spans the viewport. It paints no background tint, so the board stays fully visible next to the panel and looks usable. Every pointer event over the board lands on that wrapper instead, so while the panel is open the board cannot be scrolled horizontally and a column cannot be scrolled vertically.

`document.elementFromPoint(400, 400)`, a point well inside the visible board, returns the wrapper rather than the card under the pointer.

The wrapper is also what carries the click to close behaviour (`e.target === e.currentTarget`), so the same element that blocks the board is the one that dismisses the panel.

Source: `apps/web/src/features/issue/components/detail/IssueDetail.tsx`, the wrapper `div` and its `onClick`.

The three sibling panels (`GodProjectDetailPanel`, `GodUserDetailPanel`, `RoleEditorPanel`) use the same wrapper but paint `bg-black/20`, so there the blocking is visible and intended. The issue panel is the one case where nothing signals it.

### Steps to reproduce

1. Open a project whose board is wider than the viewport and has one column holding more cards than fit vertically.
2. Click a card to open the detail panel.
3. Put the pointer over the board, clear of the panel, and scroll.
4. Nothing moves on either axis until the panel is closed.

Measured on a board with 12 cards in one column, wheel applied at 400px from the left edge with the panel open:

| Wheel | Board `scrollLeft` | Column `scrollTop` |
| --- | --- | --- |
| `deltaY` 400 | 0 | 0 |
| `deltaX` 400 | 0 | 0 |

### How are you running it

Self-hosted (Docker Compose)

### Version or commit

v0.16.0

### Environment

Chrome 151, Windows 11. Reproduced the same way against `bun run dev` on `main` at ad895a4.

### Notes

I have a fix ready and will open a pull request against this issue. It makes the wrapper `pointer-events-none` and the panel itself `pointer-events-auto`, and moves click to close onto a document listener that only reacts to a click in the same top level layer as the panel, so a click in a portalled popover, dropdown, select, dialog or toast leaves the panel open.

OthmanAdi:
## What

The issue detail panel no longer blocks the board behind it. The wrapper that spans the viewport takes no pointer events, the panel itself takes them, and click to close moves to a document listener.

## Why

The wrapper is `fixed inset-0` and paints no background tint, so the board stays fully visible next to the panel and looks usable. Every pointer event over it landed on the wrapper instead, so neither board axis scrolled until the panel was closed. `document.elementFromPoint(400, 400)`, a point well inside the visible board, returned the wrapper rather than the card under the pointer.

The three sibling panels (`GodProjectDetailPanel`, `GodUserDetailPanel`, `RoleEditorPanel`) share the wrapper but paint `bg-black/20`, so there the blocking is visible and intended. They are left as they are.

Closes #287

### On the click to close listener

The wrapper carried the dismissal (`e.target === e.currentTarget`), so making it transparent to pointer events meant moving that behaviour. `useExitOnClickOutside` sits next to the existing `useExitOnEscape` and reads the click from the document.

Two details in it are deliberate:

**It compares top-level layers, not a list of overlay selectors.** The app renders into one child of the body, and every portalled surface is a further child. `surface.closest('body > *')` is the panel's layer, so a click in a popover, dropdown, select, dialog or toast never reaches the exit. A selector allowlist was the first attempt and it had holes: `SelectContent` is pinned to `position="item-aligned"`, whose Radix path renders a bare positioned `div` with no `data-radix-popper-content-wrapper`, so every select option click inside the panel would have closed it.

**It listens for `click`, not `pointerdown`.** The title and description in the panel save on blur, and blur fires between `pointerdown` and `click`. Exiting on the earlier event unmounted the panel before either save ran and dropped the edit. `click` also means a scrollbar drag and a right click do not dismiss the panel.

## How to test

1. Open a project whose board is wider than the viewport, with one column holding more cards than fit.
2. Click a card to open the detail panel.
3. Scroll with the pointer over the board, clear of the panel. Both axes move now.
4. Edit the title, then click the board without leaving the field first. The panel closes and the new title is saved.
5. Open a dropdown or a select inside the panel and pick a value. The panel stays open.
6. Click empty board space. The panel closes. Escape still closes it.

Measured with the same wheel input at 400px from the left edge, panel open:

| Wheel | Board `scrollLeft` before | after | Column `scrollTop` before | after |
| --- | --- | --- | --- | --- |
| `deltaY` 400 | 0 | 0 | 0 | 400 |
| `deltaX` 400 | 0 | 400 | 0 | 0 |

`bun test` in `apps/web`: 121 pass, 0 fail. The five new cases fail if the hook regresses to `pointerdown` or loses its listener cleanup, both checked by reverting each change and re-running.

## Checklist

## Screenshots

Same wheel input in both, panel open. Before, the board does not move. After, it has scrolled horizontally to Todo and In Progress and the column has scrolled down.

Before

Board frozen behind the open issue detail panel

After

Board scrolled while the issue detail panel stays open

---

Thanks for building this. The panel without a backdrop tint is the right call, and the codebase made this easy to trace: `AGENTS.md` in the root and in `apps/web` answered every placement and style question without a round trip, and `useExitOnEscape` gave the new hook its shape.

### DIFF
--- a/apps/web/src/features/issue/components/detail/IssueDetail.tsx
+++ b/apps/web/src/features/issue/components/detail/IssueDetail.tsx
@@ -1,9 +1,10 @@
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 import { Maximize2, X } from 'lucide-react';
 import { type ProjectDetail, type IssueDetail as IssueDetailRow } from '@/lib/api';
 import IssueDetailContent from './IssueDetailContent';
 import IssueActionsBar from '../actions/IssueActionsBar';
 import { useExitOnEscape } from '@/hooks/useExitOnEscape';
+import { useExitOnClickOutside } from '../../hooks/useExitOnClickOutside';
 import { Button } from '@/components/ui/button';
 import { useTranslations } from 'next-intl';
 
@@ -25,19 +26,23 @@ export default function IssueDetail({
   const t = useTranslations('issue');
   const tCommon = useTranslations('common');
   const [issue, setIssue] = useState<IssueDetailRow | null>(null);
+  const panelRef = useRef<HTMLDivElement>(null);
 
   useExitOnEscape(onClose);
+  useExitOnClickOutside(panelRef, onClose);
 
   return (
     <div
       // Above the floating chat button (z-40), which sits in the corner the panel's
       // last-comment bubble uses.
-      className="fixed inset-0 z-50 flex"
-      onClick={(e) => e.target === e.currentTarget && onClose()}
+      className="pointer-events-none fixed inset-0 z-50 flex"
     >
       {/* No padding at the top: the sticky header carries it, so it can sit flush
           against the panel edge with nothing showing above it. */}
-      <div className="ms-auto flex h-full w-full flex-col overflow-y-auto border-s bg-card px-5 pt-0 pb-5 sm:w-[720px] sm:max-w-[92vw] sm:px-8">
+      <div
+        ref={panelRef}
+        className="pointer-events-auto ms-auto flex h-full w-full flex-col overflow-y-auto border-s bg-card px-5 pt-0 pb-5 sm:w-[720px] sm:max-w-[92vw] sm:px-8"
+      >
         {/* The header stays at the top while the body scrolls under it. Negative
             margins cancel the panel padding so its translucent, blurred backdrop
             spans the full panel width. */}
--- a/apps/web/src/features/issue/hooks/useExitOnClickOutside.test.tsx
+++ b/apps/web/src/features/issue/hooks/useExitOnClickOutside.test.tsx
@@ -1,0 +1,152 @@
+import assert from 'node:assert/strict';
+import { afterEach, beforeEach, describe, it } from 'node:test';
+import { act, useRef } from 'react';
+import type { Root } from 'react-dom/client';
+import { JSDOM } from 'jsdom';
+import { useExitOnClickOutside } from './useExitOnClickOutside';
+
+const replacedGlobals = [
+  'window',
+  'document',
+  'navigator',
+  'Element',
+  'HTMLElement',
+  'Event',
+  'MouseEvent',
+  'IS_REACT_ACT_ENVIRONMENT',
+] as const;
+
+// The app renders into one child of the body, so the surface and the page behind it
+// share a layer. An overlay is appended as a further child of the body.
+const markup =
+  '<!doctype html><div data-testid="app"><div id="root"></div>' +
+  '<div data-testid="page">Board</div></div>';
+
+let dom: JSDOM;
+let root: Root;
+let exits: number;
+let documentClickListeners: number;
+let originalGlobalDescriptors: Map<string, PropertyDescriptor | undefined>;
+
+function Surface() {
+  const ref = useRef<HTMLDivElement>(null);
+  useExitOnClickOutside(ref, () => {
+    exits += 1;
+  });
+
+  return (
+    <div ref={ref} data-testid="surface">
+      <button type="button" data-testid="inside">
+        Inside
+      </button>
+    </div>
+  );
+}
+
+function render() {
+  act(() => root.render(<Surface />));
+}
+
+function element(testId: string): HTMLElement {
+  const found = document.querySelector<HTMLElement>(`[data-testid="${testId}"]`);
+  assert.ok(found);
+  return found;
+}
+
+function dispatch(target: HTMLElement, type: string) {
+  act(() => target.dispatchEvent(new window.MouseEvent(type, { bubbles: true })));
+}
+
+function appendOverlayLayer(): HTMLElement {
+  const layer = document.createElement('div');
+  const target = document.createElement('button');
+  target.type = 'button';
+  layer.append(target);
+  document.body.append(layer);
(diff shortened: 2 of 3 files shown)


OthmanAdi:

OthmanAdi:
Pushed a second commit. I audited my own change before asking anyone else to read it, and it turned out to be wrong in a way the first version hid.

**What was broken.** A click reports the common ancestor of press and release. Selecting text in the panel and releasing the mouse over the board therefore reported an element outside the panel, and the panel closed mid-selection. Copying text out of a description is an ordinary thing to do, so this would have shipped as a new annoyance in place of the old one. Confirmed against a running build before changing anything, and again after.

**The fix.** The press now decides. A gesture that starts inside the panel never exits. Everything the click listener already got right is unchanged, because it still exits on a click whose press landed outside.

**Why click and not pointerdown.** Worth recording, because pointerdown looks more responsive and is the obvious thing to reach for later. The title and description save on blur, and blur fires between pointerdown and click. Exiting on the earlier event unmounts the panel before either save runs and drops the edit. `click` also means a scrollbar drag and a right click raise nothing at all.

**Why layers and not a selector list.** The first version excluded overlays by naming their selectors, and that had a real hole: `SelectContent` is pinned to `position="item-aligned"`, whose Radix path renders a bare positioned div with no popper wrapper attribute, so every select option click inside the panel would have closed it. Comparing top-level layers covers every portal without naming any. The constraint that makes it correct is now written down in the hook.

**Verification, since this is behaviour that is easy to assert and hard to prove.**

| Case | Result |
| --- | --- |
| Wheel over the board, both axes, panel open | scrolls, panel stays |
| Selection dragged from the panel onto the board | panel stays |
| Click a select option inside the panel | panel stays |
| Click empty board space | panel closes |
| Edit the title, then click the board without leaving the field | closes, and the edit is in the database after a reload |

Seven unit cases, and I checked each one goes red for the right reason rather than trusting that it would: removing the listener cleanup fails the cleanup case, switching back to pointerdown fails three, and removing the press gate fails exactly the two that cover it.

`bun run typecheck`, `bun run lint` and `bun run format:check` pass. `bun test` in `apps/web` is 123 pass, 0 fail.

One thing I deliberately left alone: `GodProjectDetailPanel`, `GodUserDetailPanel` and `RoleEditorPanel` share this wrapper but paint `bg-black/20`. There the blocking is visible and intended, so changing them would be a different decision, not this fix.

croffasia (approved):

croffasia:
@OthmanAdi Nice work! Thanks for the PR - merged 🎉

OthmanAdi:
Two more corrections, both from auditing my own diff rather than from anyone reporting them.

**A click that raises no press was judged on the previous gesture.** The press flag stayed set until the next press, so after any press inside the panel a keyboard Enter or a synthetic click on the board left the panel open. Reading and clearing the flag on each click keeps the selection-drag behaviour and judges those activations on their own target. There is a case for it now, and it goes red with the old latching flag.

**I overclaimed in the first commit body.** It said the board's cards stay clickable. They do not, in effect. Next hydrates at `document`, so React's delegated listener is registered there before this hook's and runs first: clicking a card sets the open issue, then the document handler clears it, React batches both, and the panel closes. I checked this against a running build rather than reasoning about it, and the panel closes rather than switching to the card.

That result is the same as on main, where the invisible layer swallowed the click entirely, so it is not a regression. But the claim was wrong and I would rather say so than leave it in the pull request. Making it true would mean `preventDefault` in the board card plus a `defaultPrevented` check here, which widens the diff into a component that has nothing to do with the scroll bug. If you would rather have that behaviour, say so and I will send it separately.

**Comment discipline.** The block violated two rules in `AGENTS.md`: it noted what the surface covers and how the page behind it behaves, which the layout and styling rule forbids, and it corrected one event to another in the "not X" shape the writing style forbids. Rewritten to state the current design only.

Current state of the branch: `bun run typecheck`, `bun run lint` and `bun run format:check` pass, and `bun test` in `apps/web` is green with eight cases on this hook. Each was checked to fail for the right reason by reverting the specific line it covers: the cleanup, the event choice, the press gate, and now the read-and-clear.

OthmanAdi:
Thanks for merging. One correction landed on the branch just after the merge, so it is not in this one: #300.

Auditing my own diff afterwards, the press flag latches. It is only reassigned by the next press, so after any press inside the panel an activation that raises none inherits the previous gesture's answer, and a keyboard Enter on a board control leaves the panel open. Escape still closes it, so it degrades keyboard use rather than trapping anyone, but the cases here did not cover that path. #300 reads and clears the flag on each click, keeps the selection-drag behaviour, and adds the case.

Also worth retracting from the description above: it said the board's cards stay clickable. They do not, in effect. Next hydrates at `document`, so React's delegated listener runs before this hook's; clicking a card sets the open issue and the document handler then clears it in the same batch, so the panel closes rather than switching. I checked that against a running build. The result matches the behaviour before this change, where the invisible layer swallowed the click entirely, so nothing regressed, but the claim was wrong.

PR #291, opened 2026-09-04
fix(api): apply the instance mcp default when a project is copied

OthmanAdi:
## What

A copied project takes `mcpEnabled` from the instance default in god mode, the same as a project created from scratch.

## Why

`createProject` reads `getProjectDefaults()` and writes `mcpEnabled` from it. `copyProject` sets every other feature flag from the source but never sets `mcpEnabled` at all, so the insert falls through to the column default, which is `false`.

The result is that the god mode setting added in #227 is silently ignored on this path. An instance whose default is on produces a copy that no agent can reach, and the toggle has to be repeated by hand on every copy. A template project is exactly the thing people copy, so this is the common case rather than an edge one.

Reproduced against v0.16.0, with no stored override so the instance default is on:

| Source `mcpEnabled` | Instance default | Copy gets |
| --- | --- | --- |
| `false` | `true` | `false` |
| `true` | `true` | `false` |

The second row is the one that shows what is happening: the copy takes neither the source value nor the instance default.

The source project's own value is still not carried, which was the intent of the original comment. A template with MCP switched off should not decide it for its copies. What changed is where the copy gets the value instead.

## How to test

1. In god mode, leave the new-project MCP default on.
2. Create a project and turn MCP off in its settings.
3. Copy it.
4. The copy is reachable over MCP, matching what a newly created project does.

Turning the instance default off and repeating gives a copy with MCP off.

## Checklist

The new case sets the source to MCP off and asserts the copy comes back on, so it fails on the current behaviour rather than passing either way. Full suite through the Docker gate:

```
docker compose -f docker-compose.test.yml build
1485 pass, 0 fail, 86 files
```

## Screenshots

Not a UI change.

---

Thanks for the instance default, it is the right shape for an instance driven through MCP. This is just the one creation path it does not reach yet.

### DIFF
--- a/apps/api/src/modules/projects/__tests__/integration/projects.test.ts
+++ b/apps/api/src/modules/projects/__tests__/integration/projects.test.ts
@@ -339,6 +339,19 @@ describe('projects', () => {
       expect(settings.data?.mcpEnabled).toBe(true);
     });
 
+    // The other direction, so the copy is shown to read the default rather than to
+    // carry a fixed value. The first account of a fresh database holds the god role.
+    it('copies a project with MCP off once the instance default is turned off', async () => {
+      const { api } = await signUpClient();
+      await api.projects.post({ key: 'SRC', name: 'Source' });
+      await api.god['project-defaults'].put({ mcpEnabled: false });
+
+      await api.projects({ projectKey: 'SRC' }).copy.post({ key: 'DST', name: 'Destination' });
+
+      const settings = await api.projects({ projectKey: 'DST' }).settings.get();
+      expect(settings.data?.mcpEnabled).toBe(false);
+    });
+
     it('copies the estimate kinds and time logging the source project carries', async () => {
       const { api } = await signUpClient();
       await api.projects.post({ key: 'SRC', name: 'Source' });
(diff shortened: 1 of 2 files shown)


OthmanAdi:
recheck

OthmanAdi:
Adding the receipt I owed on the test.

I had claimed the new case fails on the current behaviour rather than passing either way, but I had reasoned it rather than run it. So I reverted the one-line fix in the working tree and ran the full Docker gate against it:

```
docker compose -f docker-compose.test.yml build api-test

(fail) projects > copy > takes mcpEnabled from the instance default, not from the source project
1484 pass, 1 fail
Ran 1485 tests across 86 files. [471.66s]
```

Exactly one failure, and it is the new case. Nothing else in the suite depends on the change. With the fix in place the same gate is 1485 pass, 0 fail.

The behaviour itself was reproduced against a running v0.16.0 before writing anything, with the instance default left on and no stored override:

| Source `mcpEnabled` | Copy gets |
| --- | --- |
| `false` | `false` |
| `true` | `false` |

The second row is the one that identifies the cause. The copy takes neither the source value nor the instance default, so it is falling through to the column default rather than carrying anything.

OthmanAdi:
Added the other direction, on review. The first case alone is satisfied by a fixed `true` in the copy, since the column default is `false`. Turning the instance default off in god mode and copying again shows the value is read rather than written, which is the pair the create path already carries in `god-project-defaults.test.ts`.

croffasia (approved):

croffasia:
Thanks @OthmanAdi 
Merged

Issue #258, opened 2026-08-30
[feature]: Add workspaces to group projects and share members across them


shiouen:
### Problem

Every project manages its own members independently. Adding someone to multiple projects means creating a separate invite for each one. For instances with several projects and overlapping teams, this creates repetitive work and makes it easy to forget a project when onboarding or offboarding someone.

SCIM groups partially solve this (an identity provider can map a group to multiple projects), but that requires an external IdP — there is no in-app grouping above the project level.

### Proposed behaviour

  Introduce workspaces as an optional layer between the instance and its projects.

  - A workspace groups projects and has its own member list with a role (owner / member).
  - A workspace member gains access to every project in the workspace, with a configurable default project role (or a per-project override).
  - Projects can still exist outside any workspace, keeping today's per-project invite flow for standalone use.
  - Inviting someone to a workspace adds them to all its projects at once; removing them removes them from all of them.
  - The existing per-project invite and SCIM paths remain: a project inside a workspace can still have direct members who are not workspace members, while it inherits all the users in the workspace.

  - Per-project roles and permissions are unchanged — the workspace controls access, the project role controls what you can do.
  - Projects without a workspace work exactly as they do today.
  - SCIM provisioning remains a separate path; a SCIM group can map to a workspace instead of (or in addition to) individual projects.
  - God mode is instance-wide and unrelated to workspaces.

Plane already has this concept.

### Alternatives you considered

Bulk management of user invitations to projects instead of one per one - but that sends out a lot of invites

  - Tags / groups on projects: lighter than workspaces but doesn't solve the membership problem, only organization.
  - Instance-wide roles: too coarse; the point of per-project permissions is that not everyone needs the same access everywhere.
  - Relying solely on SCIM groups: works for organizations with an IdP, but self-hosted instances without one get no grouping at all.

### Contribution

- [x] I am willing to open a pull request for this

croffasia:
Thanks for the detailed write-up.

I’m actually already building this - it’s called Teams, but it’s basically the same thing as workspaces. It’s in progress on the feat/teams branch, feel free to take a look and tell me if it covers what you had in mind.

Bulk invites is a good point too, I’ll keep it in mind.

shiouen:
Sure thing, I'll have a look. Might want to consider renaming to workspace - at big enterprise teams can be shifted around, merged, renamed, quite often. If they think in the concept of a workspace, for either team, product, event, ... it might help them make better choices in how they organize those groupings.

[...]

PR #292, opened 2026-09-04
docs: scope initiatives to a project in the feature list

OthmanAdi:
## What

Corrects one line in the README feature list. Initiatives group work inside a project, not across projects.

## Why

The line read "Initiatives that group and track work across projects". The code says the opposite in three places:

- `packages/db/src/schema/app.ts` comments the table as "A strategic grouping of issues inside a project (project-scoped, not cross-project)".
- `initiative.project_id` is not null and there is no join table, so one initiative belongs to exactly one project.
- Every route sits under `/projects/:projectKey/initiatives`, and the id-addressed ones resolve the owning project before the permission check.

Worth correcting rather than leaving as a small inaccuracy, because cross-project grouping is a thing people ask for (#258 among others) and this line reads like it already exists.

## How to test

Read the line against the schema comment.

## Checklist

- [ ] Tests added or updated: documentation only
- [x] Docs updated

## Screenshots

Not a UI change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ be moved up the roadmap. Ask on [Telegram](https://telegram.me/croffasia).
 - Quick actions that run on an issue, and auto-assignment when an issue moves into a state
 - Freeform notes boards: sticky notes on a canvas, with colors, checklists, and connections
 - Share a view or an issue by public link, read-only and without sign-in
-- Initiatives that group and track work across projects
+- Initiatives that group and track related work inside a project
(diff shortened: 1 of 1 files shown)


OthmanAdi:
recheck

OthmanAdi:
Retitled and the commit subject rewritten. The old subject was "initiatives are project-scoped, not cross-project", which is the "not X" shape the writing style in AGENTS.md rules out for commit messages, and it stated a correction rather than the resulting state. Since the title is the squash subject, it would have carried into the history that way.

PR #295, opened 2026-09-04
fix(web): stop a description saving itself when nobody edited it

OthmanAdi:
## What

A description editor reports a blur only when the document actually changed, so opening an issue and clicking through its description no longer saves it.

## Why

Focusing a description and clicking away saved it, wrote a "changed the description" entry to the activity feed and notified its watchers, with nobody typing anything. The stored markdown changed too: a bare URL came back wrapped in angle brackets.

The cause is that the markdown round trip is not an identity. `IssueMarkdownEditor` configures `Markdown` with `linkify: true`, so a bare URL parses into a link node, and the serialiser writes that node back as an autolink in angle brackets. The blur handler hands `editor.storage.markdown.getMarkdown()` to its caller, the caller compares it against the stored text, sees a difference and patches.

That means any description containing a bare URL saves itself on every focus, forever, without ever converging, because each save stores the angle bracket form and the comparison is against a fresh parse of it.

The consequence is worse on a shared instance than it looks. Reading an issue and clicking in the description is enough to put a false edit in its history and mail every watcher about it.

Reproduced on v0.16.0 against a running instance, on a fresh issue whose description was `See https://example.com/spec for details.`, by clicking into the description, typing nothing, and clicking a heading:

| | Activity rows | Stored description |
| --- | --- | --- |
| Before | `created`, `description` | `See <https://example.com/spec> for details.` |
| After | `created` | `See https://example.com/spec for details.` |

A real edit is unaffected. Typing into the same description and blurring writes exactly one `description` row, and focusing and blurring again after that writes none.

## How the flag behaves

Set by `onUpdate`, which every ProseMirror document transaction raises, so typing, pasting, a dropped file, a table edit and a programmatic insert all count. Selection movement raises `onSelectionUpdate` instead and does not.

Cleared by the remount, not by the blur. The parent keys the editor on `issue.updatedAt`, so a successful save remounts it and the flag resets. A save that fails leaves the flag set, so the next blur reports again rather than losing the edit. That failure direction is deliberate: reporting twice costs a redundant patch, reporting never costs the user their text.

## How to test

1. Create an issue whose description contains a bare URL.
2. Open it, click into the description, type nothing, click the title or a heading.
3. No `description` entry appears in the activity feed and the stored text is unchanged.
4. Type something and click away. One entry appears and the edit is saved.

## Checklist

- [ ] Tests added or updated: the behaviour lives in a TipTap editor instance, and `apps/web` has hook tests but no component test harness. Standing up one for a bug fix looked like the wrong trade, so this is verified against a running instance with the database receipts in the table above. Happy to add the harness if you would rather have it.

`bun test` in `apps/web` is 123 pass, 0 fail.

## Screenshots

The visible symptom is an activity row, so the table above is the evidence rather than an image.

---

Found while auditing an unrelated change of mine: I dragged a text selection out of the issue panel and noticed a description write in the activity feed that I had not made.

### DIFF
--- a/apps/web/AGENTS.md
+++ b/apps/web/AGENTS.md
@@ -112,3 +112,13 @@ next-intl, language from the `NEXT_LOCALE` cookie — no `[locale]` route segmen
   image depends on them.
 - When touching `localStorage`/`window` in a render path (e.g. a `useState` initializer), guard
   with `typeof window === 'undefined'` — client components still server-render.
+
+<!-- BEGIN:nextjs-agent-rules -->
+
+# This is NOT the Next.js you know
+
+This version has breaking changes — APIs, conventions, and file structure may all differ from your training data. Read the relevant guide in `node_modules/next/dist/docs/` (resolved from this file's directory; in monorepos the `next` package may not be visible from the repo root) before writing any code. Heed deprecation notices.
+
+This block is written and re-added by `next dev` — verify at `node_modules/next/dist/server/lib/generate-agent-files.js`. Removing it from a diff only re-creates the uncommitted change; committing it with your work keeps the tree clean.
+
+<!-- END:nextjs-agent-rules -->
--- a/apps/web/src/features/issue/components/editor/IssueMarkdownEditor.test.ts
+++ b/apps/web/src/features/issue/components/editor/IssueMarkdownEditor.test.ts
@@ -3,6 +3,7 @@ import { afterEach, beforeEach, describe, it } from 'node:test';
 import { Editor } from '@tiptap/core';
 import Link from '@tiptap/extension-link';
 import StarterKit from '@tiptap/starter-kit';
+import { Markdown } from 'tiptap-markdown';
 import { JSDOM } from 'jsdom';
 import { issueEditorStarterKitOptions } from './IssueMarkdownEditor';
 import { openLinkOnModifierClick } from '../../utils/modifierClickLink';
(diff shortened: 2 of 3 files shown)


OthmanAdi:
Adding the verification for the one way this change could go wrong.

The risk in gating a blur on a dirty flag is losing a save, which is worse than the phantom save it removes. The flag is set by `onUpdate`, so the question is whether every real content change raises it. Typing obviously does. The case worth checking is a change made by a command rather than a keystroke, because that is how a dropped or pasted file gets inserted: `insertFiles` runs `insertContentAt` through an editor chain.

Tested on a running instance against a fresh issue described as `plain text with no url`:

| Step | Activity rows after | Stored description |
| --- | --- | --- |
| Open, click into the description, type nothing, click away | `created` only | unchanged |
| Select all, apply bold with the keyboard command, click away | `created`, `description` | `**plain text with no url**` |

So a command-driven transaction sets the flag and still saves, and a focus that changes nothing still saves nothing. A dropped or pasted attachment goes through the same transaction path.

The failure direction is also deliberate. The flag is cleared by the remount that a successful save causes, not by the blur itself, so a save that fails leaves it set and the next blur reports again. Reporting twice costs a redundant patch; reporting never would cost the user their text.

One thing I left out on purpose. The same round-trip instability means a real edit still stores the canonical serialisation, so editing a description that contains a bare URL will store it as an autolink. That is the serialiser being consistent once you actually edit, and changing it would mean turning off `linkify`, which would stop bare URLs being clickable. That looked like a product decision rather than part of this fix.

OthmanAdi:
Two things, one of which corrects the description above.

**The cause is more specific than I wrote.** I said `linkify` makes the round trip non-identity. Writing a test showed that is not it: with StarterKit and Markdown alone, `linkify: true` included, a bare URL survives the round trip unchanged. It is the link extension's `autolink` that parses the URL into a link node, which the serialiser then writes back in angle brackets. The test pins exactly that, and the comment in the component now says so.

**There is a test now.** `IssueMarkdownEditor.test.ts` already drives a real editor under jsdom, which I had missed. The harness needed `DOMParser` and the node constructors added for tiptap-markdown to parse content, and the case then asserts the round trip returns the angle bracket form and not the input. That pins the reason the guard exists: if a future version stops rewriting the URL, this goes red and tells whoever reads it that the guard is now dead weight.

**One coupling worth naming.** The flag is cleared by the remount, and that remount comes from the caller keying the editor on the value it stores. Both callers do that today for their own reasons. If either stopped, the flag would never clear and every focus after one real edit would save again. The comment now names both callers so that is visible rather than implicit. Clearing it inside the component instead is possible, but every version I tried introduced a worse failure: resetting on a `defaultValue` change would discard a pending edit when a live refresh arrives mid-typing. I would rather leave the coupling documented than trade a loud bug for a quiet one, though I will take a different shape if you prefer one.

**Also worth saying plainly:** this suppresses the no-edit case, it does not make the round trip an identity. One real edit on a description holding a bare URL still stores it as an autolink. Changing that would mean turning off autolink, which would stop bare URLs being clickable, and that reads like your decision rather than part of this fix.

croffasia (approved):

croffasia:
Thanks for the PR! Merged

PR #297, opened 2026-09-04
fix(api): answer 400 instead of 500 when an unauthenticated route gets malformed input

OthmanAdi:
## What

Three unauthenticated paths that answered 500 on malformed input now answer the status the repo already uses for the same shape elsewhere.

## Why

All three are reachable without a session, and all three turn a bad input into a 500 whose body carries an internal message.

**The two raw attachment routes take an unvalidated path segment.** `GET /attachments/:publicId/raw` and `GET /chat-attachments/:publicId/raw` declare no `params` schema, so the segment goes straight into an `eq()` against a Postgres `uuid` column. On v0.16.0:

```
GET /attachments/not-a-uuid/raw
500
{"error":"Failed query: select \"id\", \"public_id\", \"issue_id\", \"s3_key\", \"filename\",
 \"content_type\", \"size_bytes\", \"created_at\" from \"issue_attachment\" where ..."}
```

That is the table name, every column name and the query shape, returned to an anonymous caller, plus an entry in the error log for anyone who wants to fill it.

The repo already treats this exact shape as a defect elsewhere, and says so. `invites/model.ts` carries the comment "The token is a UUID column. Validating its format here turns a malformed token into a 400 instead of letting it reach Postgres and surface as a 500", and `share/model.ts` says the same. Both behave correctly today:

| Route | Malformed input | Well-formed unknown |
| --- | --- | --- |
| `/share/issue/:token` | 400 | 404 |
| `/invites/:token` | 400 | 404 |
| `/attachments/:publicId/raw` | **500** | 404 |
| `/chat-attachments/:publicId/raw` | **500** | 404 |

Only the two raw routes were missed. This applies the same schema they use.

**The server view filter engine dropped a guard its own header requires.** `views/filters.ts` opens with "Keep the matching semantics in sync with the web util", and the web util's `isEffectiveCondition` reads `Array.isArray(cond.values) && cond.values.length > 0` with the comment "`values` is checked for being a list because the whole set comes from a jsonb blob the server stores without inspecting it". The server copy has only `cond.values.length > 0`.

A view's `filters` really is stored uninspected (`t.Any()`, written verbatim), so a condition without `values` can be saved through the API or over MCP. The signed-in board renders fine because the client guard drops the condition. The public link does not:

```
GET /share/view/<token>
500
{"error":"undefined is not an object (evaluating 'cond.values.length')"}
```

Reproduced on a running v0.16.0 by saving `{"conditions":[{"field":"status","op":"is"}]}` on a shared view. Every visitor to that link sees the board-unavailable screen with nothing to tell them why, and the owner sees a working board.

## How to test

```
curl -i localhost:3000/attachments/not-a-uuid/raw          # 400, was 500
curl -i localhost:3000/chat-attachments/not-a-uuid/raw      # 400, was 500
```

For the filter: create a view, save `{"conditions":[{"field":"status","op":"is"}]}` as its filters, share it, and open the public link. It serves the board.

## Checklist

Two cases added, each next to the existing case it parallels: a non-uuid `publicId` on the public raw route, and a shared view whose condition carries no values. The existing attachment test only probed a well-formed unknown id, which is why nothing caught this.

Full suite through the Docker gate on this branch:

```
docker compose -f docker-compose.test.yml build api-test
1486 pass, 0 fail
Ran 1486 tests across 86 files. [407.22s]
```

## Screenshots

Not a UI change.

---

Found by sweeping the surfaces that answer without a session. Both fixes are the pattern already in the repo rather than a new mechanism: the params schema is the one `invites` and `share` use, and the guard is the line the web util already has.

### DIFF
--- a/apps/api/src/modules/attachments/__tests__/integration/attachments.test.ts
+++ b/apps/api/src/modules/attachments/__tests__/integration/attachments.test.ts
@@ -242,6 +242,13 @@ describe('attachments', () => {
         .raw.get();
       expect(res.status).toBe(404);
     });
+
+    // The route is public, so a malformed id reaching Postgres would answer 500 with
+    // the driver's message to an anonymous caller.
+    it('returns 400 for a publicId that is not a uuid', async () => {
+      const res = await api.attachments({ publicId: 'not-a-uuid' }).raw.get();
+      expect(res.status).toBe(400);
+    });
   });
 
   describe('import (url or base64)', () => {
--- a/apps/api/src/modules/attachments/index.ts
+++ b/apps/api/src/modules/attachments/index.ts
@@ -20,6 +20,7 @@ import {
   AttachmentListResponse,
   importAttachmentBody,
   issueParams,
+  publicIdParams,
   rawAttachmentQuery,
   uploadAttachmentBody,
 } from './model';
@@ -374,10 +375,11 @@ export const attachmentRoutes = new Elysia({
       return new Response(obj.body, { headers });
     },
     {
+      params: publicIdParams,
       query: rawAttachmentQuery,
       // Public route: no 401/403. Returns a raw Response (bytes), so no typed 200
-      // body — Elysia cannot validate a raw Response. Only the 404 it can throw.
-      response: { ...errors(404) },
+      // body — Elysia cannot validate a raw Response. Only the statuses it can throw.
+      response: { ...errors(400, 404) },
       detail: { summary: 'Download or preview an attachment (public, no auth)' },
     },
   );
--- a/apps/api/src/modules/attachments/model.ts
+++ b/apps/api/src/modules/attachments/model.ts
@@ -14,6 +14,10 @@ export const AttachmentListResponse = t.Array(AttachmentResponse);
 
 export const issueParams = t.Object({ issueId: t.Numeric() });
 
+// The public id is a UUID column. Validating its format here turns a malformed id
+// into a 400 instead of letting it reach Postgres and surface as a 500.
+export const publicIdParams = t.Object({ publicId: t.String({ format: 'uuid' }) });
+
 export const uploadAttachmentBody = t.Object({ file: t.File() });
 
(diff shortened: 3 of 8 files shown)


OthmanAdi:
Two corrections and one thing this pull request does not close.

**It does not close the mechanism.** `planner.ts` returns the error's own message in every 500 body, and drizzle's message is the failed statement. So this fixes the two inputs I found reaching Postgres unchecked, and the next unvalidated one on any public route discloses the same way. That is a behaviour change on every route and the shape is the maintainer's call, so it is #299 rather than a patch here. Worth reading the two together: with #299 done, these uuid guards become defence in depth instead of the whole fix.

**Corrections pushed since opening.**

The params schema is no longer declared twice. `chat-attachments/model.ts` already re-exported `rawAttachmentQuery` from the attachments model, so `publicIdParams` now comes from the same line rather than being a second identical declaration with the same comment.

The chat-attachments raw route now has the 400 case its twin had, per the coverage rule in `apps/api/AGENTS.md`.

The shared view case now asserts its own precondition. It patched a filter set through `filters: t.Any()` and never checked the patch landed, so tightening that schema later would have left the case green while testing nothing.

**On the two tests going red for the right reason.** The attachment case fails on main because the handler queries the uuid column directly. The share case fails on main because `isEffectiveCondition` throws on `undefined.length`, and the helper it uses creates the view with no filters, so the assertions are reached rather than short-circuited.

**One thing I checked rather than assumed.** `format: 'uuid'` does reject in this setup: `invites.test.ts` and `share.test.ts` already assert 400 for a non-uuid token and are green on main, so the pattern is proven here rather than borrowed from the TypeBox docs. The third public bytes route, `/avatars/:id/raw`, is deliberately out of scope: it reads from S3 and 404s inside a catch, so it never reaches Postgres.

croffasia (approved):

croffasia:
Thanks for the PR 🙏

PR #301, opened 2026-09-04
fix(web): open the issue a click asks for instead of closing the panel

OthmanAdi:
## What

Clicking a card while the issue detail panel is open shows that issue, instead of closing the panel and needing a second click.

## Why

Reported by a user of a self-hosted instance: reading a second issue from the board took two clicks, one to close and one to open, with no way to tell why the first one only closed.

The panel reads a click outside itself as a request to close. A click on a card is outside it, so both things run for one click: the card sets the open issue, the panel's document listener clears it. React handles the card's click first, being a discrete event flushed before the document listener runs, so the close lands on top of the open and the panel ends closed.

Opening an issue now marks its click handled with `preventDefault`, and the panel leaves a handled click alone. That is the convention already in `useExitOnEscape`, which skips a key another surface has handled.

A click on the board that opens nothing is unaffected and still closes the panel.

## Scope

Every view that can open an issue from the page behind the panel: the kanban card, the calendar chip, the calendar overflow list, the unscheduled card, and the timeline row, bar, link row and subtask row. Leaving one of them out would have made the behaviour depend on which view you happened to be in.

## How to test

1. Open an issue from the board.
2. Click a different card once. The panel shows that issue.
3. Click empty space in a column. The panel closes.
4. Repeat in the calendar and timeline views.

Verified against a running instance on all three checks, and the first two are the ones that changed.

## Checklist

- [ ] Tests added or updated: the behaviour is the interaction between a React click handler and a document listener, and `apps/web` has hook tests but no component harness that renders the board. Standing one up for this looked like the wrong trade, so it is verified against a running instance. The hook's own change is covered by the existing cases, which still pass. Happy to add the harness if you would rather have it.

`bun test` in `apps/web` is 124 pass, 0 fail.

## Screenshots

The change is a state transition rather than a visual one, so the steps above are the evidence.

---

This is the behaviour I described as out of scope in #288 and offered to send separately. A user hit it within an hour of that release going onto their instance, so here it is.


OthmanAdi:
Second commit: the card the panel is showing now closes it when clicked again.

Making a card click replace what the panel shows left the card it was opened from with nothing to do, so the panel could only be dismissed from its close button, Escape, or empty board space. Clicking it again is the obvious way to put it away and it did nothing.

Every view opens through the one callback in the Shell, so the toggle sits there rather than in each view.

Checked against a running instance, all six together:

| Action | Result |
| --- | --- |
| Click a card | Opens it |
| Click the same card again | Closes |
| Click a different card | Switches to it |
| Click empty board space | Closes |
| Wheel over a column, panel open | Column scrolls, panel stays |
| Wheel across, panel open | Board scrolls, panel stays |

`bun run typecheck`, `bun run lint` and `bun run format:check` pass; `bun test` in `apps/web` is 124 pass, 0 fail.

croffasia (approved):

croffasia:
Merged! A few spots are still not covered though: table view - tasks, links and subtasks. And on the kanban board - subtasks and links on the card.

PR #259, opened 2026-08-30
feat: add native Git connections and development status

grootbro:
## What

- Add token-based GitHub, GitLab, Gitea, Forgejo, and Bitbucket provider connections with repository discovery and managed webhooks.
- Add an issue Development panel for branches, pull request state, GitHub check runs, and GitLab pipeline status.
- Link issues as soon as a branch such as `KEY-123-short-title` is created, then promote that branch link to the pull request without losing CI history.
- Let `Skip KEY-123` and `Ignore KEY-123` remove an existing link on a later pull request edit.
- Add optional provider-account backlink comments, enabled by default.
- Add actionable provider errors for invalid tokens, organization SSO, rate limits, permissions, missing repositories, and rejected webhook settings.
- Add copyable `gh` and `glab` webhook commands and provider-specific managed webhook subscriptions.
- Restructure repository settings into enablement, webhook endpoint, recommended provider connections, manual setup, automations, and recent deliveries.

## Why

The provider-neutral webhook receiver can already move issues, but repository setup is manual and issue details do not show linked development state or CI results. This implements the repository mapping, branch and pull-request linking, and CI-status portions of #162 while keeping the broader issue open for full issue/comment synchronization and agent-created repositories.

Refs #162

## How to test

1. Enable repository integration in a project.
2. Connect a provider token and select a repository with webhook administration permission.
3. Create a branch named `KEY-123-short-title` and confirm that the issue immediately shows a branch in Development.
4. Open a pull request from that branch and confirm that the card becomes a pull request while an existing CI result remains attached.
5. Add `Skip KEY-123` to the pull request description and confirm that the link disappears after the edit webhook.
6. Run GitHub checks or a GitLab pipeline and confirm that the current status updates without reloading.
7. Toggle backlink comments, unlink a development item, and disconnect the repository; confirm that the managed webhook is removed.

## Checklist

- [x] API, Web, and DB typecheck pass
- [x] API, Web, and DB lint pass
- [x] Focused provider, parser, and CLI tests pass
- [x] Database-backed regression tests added for branch linking, skip removal, and branch-to-PR CI preservation
- [x] Database migration generated with a backfill for existing pull-request links
- [x] No new environment variables
- [x] Setup guidance is included in the repository settings UI

## Screenshots

The feature was exercised on a reference instance. Its repository names are private, so they are not included in this public PR.

PR #317, opened 2026-09-06
fix(net): allow operators to trust named hosts that resolve privately

### LINKED ISSUE #318: [bug]: self-hosted instance cannot connect to a self-hosted repository host (SSRF guard rejects private addresses)
### What happened

A self-hosted instance cannot connect to a self-hosted repository host.

Adding a Forgejo connection (Project -> Integrations -> Provider connections -> Connect Forgejo) fails at `POST /projects/:projectKey/settings/git/connections`:

```
HTTP/1.1 400 Bad Request
{"error":"url must not point to a private or local address"}
```

The forge is reachable from the api container and the token is valid: from inside the pod, `GET https://forgejo.example.internal/api/v1/version` returns 200, and the same token answers 200 on `/user`, `/user/repos`, and `/repos/{owner}/{repo}/hooks`. Only the connection form refuses.

The cause is the SSRF guard added in #313. It resolves the base URL and rejects it if any resolved address is private. On a self-hosted instance the forge normally is on a private network, and split-horizon DNS means the name resolves internally to a private address in both families:

```
10.3.60.19              -> a === 10
fd9a:d56f:8111:60::19   -> fc00::/7 unique-local
```

The guard is correct for a URL that arrives as content. A webhook target and an imported attachment come from whoever can reach the form, and the advisories behind #313 show what that reaches. A repository host's base URL is different: it is configuration, entered behind `integrations:edit` by someone describing infrastructure they already administer.

There is no way to express that today. `devRelaxed` keys off `NODE_ENV`, so the only workaround is moving `NODE_ENV` off `production`, which disables the guard for every call site and changes how the app runs.

This is a regression: native Git connections (#259) worked against a private forge before the guard landed in #313.

### Steps to reproduce

1. Run an instance where the repository host resolves to a private address (any self-hosted Gitea/Forgejo/GitLab on the same network).
2. Create an access token on that host with permission to administer repository hooks.
3. Project -> Integrations -> Provider connections -> Connect Forgejo.
4. Enter the provider URL and the token, and press Connect.
5. The request returns 400 `url must not point to a private or local address`.

A public hostname that resolves privately reproduces it identically, so `localtest.me` works as a local repro without a forge.

### How are you running it

Self-hosted (Docker Compose)

### Version or commit

0.17.0

### Environment

Self-hosted on Kubernetes, api and web on separate origins. The repository host is a self-hosted Forgejo 16.0.3 on the same private network, reached over https with a valid publicly-issued certificate.

### Logs

The 400 above is the whole signal; the guard throws `UrlNotAllowedError` before any request is made, so there is no outbound attempt to log.

coffeehorn:
## What

Adds `SSRF_ALLOWED_HOSTS`, an operator-set list of hosts `@repo/net` will fetch even though they resolve to a private address. Empty by default, so no existing deployment changes behaviour.

For a self-hosted instance it goes in `.env` beside the other api and worker variables, and takes effect on restart:

```
SSRF_ALLOWED_HOSTS=git.internal.example.com
```

Three call sites share the guard — webhook delivery, attachment import, and the git provider client — so a named host is trusted by all three. Scoping it to the provider client alone would be a small change if you would rather it were narrower; see the note at the end.

## Why

Closes #318 — a self-hosted instance cannot connect to a self-hosted repository host. Adding a connection whose base URL resolves to a private address fails at `POST /projects/:projectKey/settings/git/connections` with `url must not point to a private or local address`.

The guard is right about a URL that arrives as content. A webhook target and an imported attachment are supplied by whoever can reach the form, and the advisories behind #313 show exactly what that reaches — including a response body readable back through `GET /webhooks/:webhookId/deliveries`.

A repository host's base URL is not that. It is configuration, entered behind `integrations:edit` by someone describing infrastructure they already administer, and on a self-hosted instance it is normally on the same private network as the instance itself. Both flow through one guard, which is the right design, but they do not deserve one verdict — and today there is no way to say so. `NODE_ENV` is the only lever, and moving it off `production` to configure an integration disables the guard everywhere and changes how the app runs.

Typed `fix` rather than `feat` because the behaviour it restores worked before #313: native Git connections (#259) could reach a private forge until the guard landed. Happy to retitle if you would rather the new variable read as a capability.

Gitea has `ALLOWED_HOST_LIST` and GitLab an equivalent setting for the same reason.

## How to test

The suite:

```bash
bun install
NODE_ENV=test bun test packages/net
```

To watch the behaviour change by hand, `localtest.me` resolves to `127.0.0.1`, so it reproduces without a forge. `NODE_ENV=test` matters — without it `devRelaxed` is true on a dev checkout and every URL below is admitted, so the demo would show nothing:

```bash
NODE_ENV=test bun -e '
const { assertPublicHttpUrl } = await import("./packages/net/src/index.ts");
const check = async (label, url) => {
  try { await assertPublicHttpUrl(url); console.log(`${label.padEnd(20)} allowed`); }
  catch (e) { console.log(`${label.padEnd(20)} refused — ${e.message}`); }
};
await check("unset", "https://localtest.me/");
process.env.SSRF_ALLOWED_HOSTS = "localtest.me";
await check("named", "https://localtest.me/");
await check("subdomain", "https://sub.localtest.me/");
await check("http", "http://localtest.me/");
'
```

```
unset                refused — url must not point to a private or local address
named                allowed
subdomain            refused — url must not point to a private or local address
http                 refused — url must use https
```

Deliberately narrow, and the tests assert each of these:

- empty or unset exempts nothing
- exact hostname match only — `sub.localtest.me` is refused while `localtest.me` is named
- naming one host does not admit another private host
- `https` is still required for a named host
- the resolved address is still pinned, so a named host cannot be used for DNS rebinding

### Results

```
bun run format:check   All matched files use Prettier code style!
bun run lint           11 successful, 11 total
bun run typecheck      11 successful, 11 total
NODE_ENV=test bun test packages/net   14 pass, 1 fail
```

The one failure is `pinnedFetch > reaches a public host`, which fetches `https://example.com/` with `timeoutMs: 15_000` against bun's 5s default per-test timeout. It fails identically on unmodified `upstream/main` here (8 pass, 1 fail), so it is environmental to my machine rather than caused by this change. All six new tests pass.

## Checklist

- [ ] Database schema changed: migration generated with `bun run db:generate` and committed — n/a
- [ ] Docs updated (`README.md` or the relevant `AGENTS.md`) — `packages/net` has no `AGENTS.md`; the variable is documented in `.env.example` and at the function that reads it. Happy to add a note to `docs/self-hosting.md` if you would like one.

## Screenshots

n/a — no UI change.

## Open questions

Two things I would rather you decide than guess at:

- **Scope.** The allowlist currently applies to all three call sites. Restricting it to the git provider client would keep webhook delivery and attachment import strictly public-only, at the cost of an operator not being able to point a webhook at their own internal receiver. Say which you prefer and I will adjust.
- **Name.** `SSRF_ALLOWED_HOSTS` follows Gitea's `ALLOWED_HOST_LIST` in spirit. If you would rather it read as configuration than as a security control — `TRUSTED_HOSTS`, `ALLOWED_PRIVATE_HOSTS` — that is a rename away.


coffeehorn:

croffasia:
@coffeehorn Thanks

1. Did you verify this on your own instance - does the Forgejo connection now succeed? The test plan only covers the guard check.
2. Please add the variable to the api and worker services in all three compose files.
3. Please note in .env.example that a named host is also reachable from webhook delivery and attachment import.

coffeehorn:
@croffasia I verified on my instance. The important part first, because it explains why the original test plan stopped at the guard: **this only reproduces where the name resolves privately.** On my workstation my forgejo instance resolves to a public address, so a local test passes with or without the change and proves nothing. Inside my cluster the same name resolves to an RFC1918 address, and that is where the guard fires.

So I ran the patched `packages/net` in a pod on the same network as the api, with the api's DNS config, against the live Forgejo:

```
resolved        10.3.60.19  fd9a:...:60::19
no allow-list   refused: url must not point to a private or local address
allow-list set  accepted: https://forgejo.<domain>
pinnedFetch     /api/v1/version           -> 200 {"version":"16.0.3+gitea-1.22.0"}
pinnedFetch     /api/v1/user/repos (no token) -> 401
other host      still refused
```

The 401 is the end-to-end part. With the host allowed, the exact path `connections-provider.ts` takes -- `assertPublicHttpUrl` to normalise the base URL, then `pinnedFetch` per request -- reaches Forgejo's own auth layer. Only the credential is absent; the guard is no longer in the way. The last line is the negative control: naming a different host leaves this one refused, so it is not a blanket disable.

Before that I confirmed raw egress from the api pod already worked (200 from `/api/v1/version` with plain `fetch`), so the guard was the only blocker rather than the network.

**Compose** -- done, api and worker in all three. Worth flagging that this was a real bug and not a tidy-up: none of the services use `env_file`, so an operator who set `SSRF_ALLOWED_HOSTS` in `.env` following the example would have found it had no effect at all. `docker-compose.coolify-images.yml` uses list-style `environment`, so its entry matches that form.

**.env.example** -- done. It now states that a named host is trusted by every outbound path, not only the repository connection: it also becomes a permitted webhook target -- validated by the api on create/update, and delivered to by the worker -- and a permitted source for attachment import, reachable by anyone who can create a webhook or import an attachment. It also says to set it on both processes, since the api vets and the worker delivers, and a list on only one silently covers half the paths.

croffasia (approved):

croffasia:
@coffeehorn Thanks, merged!

PR #325, opened 2026-09-07
fix: reject an issue due date before its start date

### LINKED ISSUE #324: [bug]:  An issue accepts a due date earlier than its start date
### What happened

An issue can be created with a due date that falls before its start date, for example, start Sep 8, 2026 with due date Aug 30, 2026. Both dates are saved and shown on the issue. The date pickers restrict nothing, and the API does not check the order either: create, PATCH, and bulk update all accept the pair.
 
Cycles already reject an end date before their start date, so issues and cycles disagree on the same kind of range.
 
Expected: the API rejects a due date earlier than the start date with 400, and the date pickers do not offer days that would produce the pair. A due date equal to the start date stays allowed.

### Steps to reproduce

1. Open a project
2. Click "New issue"
3. Set Start date to Sep 8, 2026.
4. Set Due date to Aug 30, 2026, earlier than the start date.
5. Create the issue and open it. Both dates are stored and displayed, no validation error appears.

The same pair is accepted through PATCH /issues/:id and through bulk update.

### How are you running it

Local development (bun run dev)

### Version or commit

257c817

### Environment

Chrome on macOS

### Logs

```shell

```

rafipopz98:
## What

An issue can no longer carry a due date earlier than its start date. The API returns 400 on create, update, and bulk update when the pair would be out of order (a due date equal to the start date stays allowed), and the start and due date calendars in the new-issue dialog and on the issue detail grey out the days that would put one date on the wrong side of the other.

## Why

Both dates could be set independently, so an issue could be saved with start Sep 8, 2026 and due Aug 30, 2026. Cycles already reject an end date before their start; issues now behave the same.

Closes #324 

## How to test

1. open a project, click "New issue".
2. Set a start date, then open the Due date calendar: days before the start date are greyed out and cannot be selected.
3. Set a due date first, then open the Start date calendar: days after the due date are greyed out.
4. Selecting the same day for both still works, and the issue saves.

## Checklist
 
## Screenshots

https://github.com/user-attachments/assets/94898c56-ca08-4f33-88aa-2d374899f43c

### DIFF
--- a/apps/api/src/modules/issues/__tests__/integration/issues.test.ts
+++ b/apps/api/src/modules/issues/__tests__/integration/issues.test.ts
@@ -301,6 +301,38 @@ describe('issues', () => {
         .issues.post({ columnId, title: 'Dated', dueDate: '2026-13-45' });
       expect(res.status).toBe(400);
     });
+
+    it('rejects a due date before the start date on create', async () => {
+      const { asOwner, columnId } = await setupProject();
+      const res = await createIssue(asOwner, columnId, {
+        startDate: '2026-09-08',
+        dueDate: '2026-08-30',
+      });
+      expect(res.status).toBe(400);
+    });
+
+    it('allows a due date equal to the start date', async () => {
+      const { asOwner, columnId } = await setupProject();
+      const res = await createIssue(asOwner, columnId, {
+        startDate: '2026-09-08',
+        dueDate: '2026-09-08',
+      });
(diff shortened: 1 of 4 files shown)


rafipopz98:

croffasia (approved):

croffasia:
@rafipopz98 Thanks 🙏 Merged

PR #348, opened 2026-09-09
fix(api): let only the creator delete a note board

egorvas:
## What

`DELETE /projects/:projectKey/note-boards/:boardId` now requires the caller to be the board's creator (`createdByUserId`), answering 403 `Only the board creator can delete the board` to anyone else who can see the board. A board with no recorded creator stays deletable by anyone holding the `note_boards.delete` permission. The route description (and so the `delete_note_board` MCP tool text) says so.

## Why

The delete route only ran `loadAccessibleBoard`, which checks that the caller may see the board. `PATCH` guards a change of visibility or granted members behind `createdByUserId`, but a member a restricted board was shared with, or any member of a public board with the delete permission, could delete the board outright. Deleting is at least as destructive as changing who sees the board, so it follows the same rule.

The exception for a `NULL` creator keeps boards created before `created_by_user_id` existed (migration 0053 only backfills a creator for owned boards) and boards whose creator's account was deleted (`ON DELETE SET NULL`) from becoming undeletable. Such a board has no owner, so it is public and already open to every member.

## How to test

1. As a project owner, create a board, make it `restricted` and grant another member.
2. As that member: `DELETE /projects/:projectKey/note-boards/:boardId` answers 403 and the board is still readable by the owner.
3. As the creator: the same request answers 204 and a following `GET` answers 404.
4. A member outside a private board gets 404 on delete, an outsider to the project 403, the same statuses `PATCH` gives them.

Integration tests: `cd apps/api && bun test --env-file=../../.env.test src/modules/note-boards` (new `delete` block, 4 tests; two of them fail on `main`).

## Checklist

### DIFF
--- a/apps/api/src/modules/note-boards/__tests__/integration/note-boards.test.ts
+++ b/apps/api/src/modules/note-boards/__tests__/integration/note-boards.test.ts
@@ -285,6 +285,54 @@ describe('note boards', () => {
     });
   });
 
+  describe('delete', () => {
+    it('lets the creator delete their board', async () => {
+      const owner = await setupOwnerProject();
+      const boardId = (await boards(owner.api).post({ name: 'Ideas' })).data!.id;
+
+      expect((await boards(owner.api)({ boardId }).delete()).status).toBe(204);
+      expect((await boards(owner.api)({ boardId }).get()).status).toBe(404);
+    });
+
+    it('keeps a granted member from deleting a board shared with them', async () => {
+      const owner = await setupOwnerProject();
+      const granted = await addMember(owner.api);
+      const boardId = (await boards(owner.api).post({ name: 'Ideas' })).data!.id;
+      await boards(owner.api)({ boardId }).patch({
+        visibility: 'restricted',
+        memberIds: [granted.userId],
+      });
+
+      const res = await boards(granted.api)({ boardId }).delete();
+      expect(res.status).toBe(403);
+      expect(res.error?.value).toEqual({ error: 'Only the board creator can delete the board' });
+      expect((await boards(owner.api)({ boardId }).get()).status).toBe(200);
+    });
+
+    it('keeps a member from deleting a public board they did not create', async () => {
+      const owner = await setupOwnerProject();
+      const member = await addMember(owner.api);
+      const boardId = (await boards(owner.api).post({ name: 'Ideas' })).data!.id;
+
+      expect((await boards(member.api)({ boardId }).delete()).status).toBe(403);
+      expect((await boards(owner.api)({ boardId }).get()).status).toBe(200);
+    });
+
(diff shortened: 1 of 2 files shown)


croffasia:
@egorvas How about we let the project owner or team owner delete other people's boards? If someone leaves the project, their board just sits there as dead weight.

egorvas:
Good catch, and it was worse than dead weight: the first version took the cleanup path away. Pushed a second commit.

Who can delete now:

- Public board: any member whose role has `note_boards.delete`, exactly as before this PR. My first commit had narrowed this to the creator, which is the regression you spotted.
- Private or restricted board: only the member who made it. A member the board was granted to gets 403 — that is the hole this PR set out to close.
- Project owner: any board, including a private one they cannot see. That is your case.
- A member who can not see a private board still gets 404, so its existence does not leak.

On "someone leaves the project": `removeMember` clears `auto_assign_user_id` on columns but does not touch note boards, so the board keeps pointing at a non-member. Deleting the account is already handled — `owner_user_id` cascades and a private board goes with it — but leaving the project is not, which is why the owner needs the power.

Two things I left for you to decide:

1. `PATCH` is still creator-only, so an owner can delete an abandoned board but not rename it or take it over. Deleting is cleanup; silently editing someone else private canvas felt like a different decision. Say the word and I will widen it the same way.
2. The route guard needs a `project_member` row, so a team owner or manager who is not in the project is refused before the board rules are reached. Including them means relaxing the guard itself. Happy to do it if you want team rank to carry here.

Verification: `bun test src/modules/note-boards` 23 pass, full api suite 1812 pass / 0 fail, `format:check`, `lint`, `typecheck` clean. Checked the new expectations fail both against my first commit (3 failures) and against `main` (2 failures, including the granted-member hole); the public-board case passes on `main`, which is how I know that behaviour is restored unchanged.

egorvas:
Withdrawing my first question: `PATCH` should stay creator-only. A private board is private precisely because nobody can take it over, and removing an abandoned one is not the same act as reading or claiming it.

Worth saying that the current code is stronger than a permission check here. `PATCH` opens with `loadAccessibleBoard`, so a private board of another member is a 404 for the project owner too: they cannot rename it or touch its canvas. And `owner_user_id` is only ever written inside the creator-gated branch, and only ever to the caller or to NULL, so there is no path that hands a board to a third party. Ownership transfer is not refused, it does not exist.

So the owner gains exactly one power over a board they cannot see: removing it. That leaves only my second question open — whether a team owner or manager without a `project_member` row should reach these routes at all, which needs the route guard relaxed, not the board rules.

croffasia (approved):

croffasia:
Agreed. We'll need to think about team owners down the line - maybe they should get the same permissions as a project owner. Merged.

PR #352, opened 2026-09-09
fix: empty the secret placeholders and warn on the published one

egorvas:
## What

- `.env.example`: `POSTGRES_PASSWORD`, `BETTER_AUTH_SECRET`, `S3_ACCESS_KEY_ID`,
  `S3_SECRET_ACCESS_KEY` and `WORKER_INTERNAL_TOKEN` are empty, like `APP_ENCRYPTION_KEY`
  already was. The `${VAR:?}` guards in `docker-compose.yml` now stop a stack that runs on the
  example file, which is what `docs/self-hosting.md` said they did.
- `@repo/crypto` gains `assertStrongSecret(name, value)` / `weakSecretReason(value)`: unset,
  shorter than 32 bytes, or one of the known example values
  (`change-me-please-generate-a-real-secret`, better-auth's built-in fallback) is refused.
  The package has no dependencies, so the worker and the bot can import it without pulling in
  the auth instance or the database client.
- `@repo/auth` runs it on `BETTER_AUTH_SECRET` at import, next to the existing `API_URL` /
  `APP_URL` checks. better-auth itself only logs a warning for a short or low-entropy secret.
- `WORKER_INTERNAL_TOKEN`: the api checks it at startup (`apps/api/src/index.ts`) and again
  per request in the new `shared/worker-token.ts`, which the three `/internal/*` route files
  now share instead of carrying a copy each. The worker checks it when `internal-api.ts`
  loads, the bot in `botConfig()`.
- `bun run setup`: Develop and Try it fill the now-empty Postgres password and MinIO keys with
  the local defaults (`itsaplan`, `minioadmin`, the values `docker-compose.dev.yml` and
  `.env.test.example` already use); Generate env walks the two MinIO keys as well.
  `EnvFile.fill(key, value)` is the primitive `generate` now builds on.
- `docker-compose.test.yml`: `worker-test` gets a token so its startup check passes.
- Docs: `docs/self-hosting.md` tells operators who kept the example values to rotate both
  secrets (sessions are signed out once; `APP_ENCRYPTION_KEY` must not change),
  `docs/development.md`, the `AGENTS.md` of api / worker / bot / auth, and the package list in
  the root `AGENTS.md` and `CONTRIBUTING.md`.

## Why

A self-hosted instance set up by following the documented steps could end up signing its
sessions with a value published in this repository. Whoever knows the signing secret can
forge the tokens better-auth verifies with it and take over any account, including the
instance owner's. The same placeholder authenticated the worker and the bot against the
`/internal/*` routes, which return the Telegram bot token and execute agent runs. Neither the
compose guards nor better-auth's own check stopped this: the guard only rejects an empty
value and better-auth only warns.

Existing installs are unaffected as long as their `.env` holds real values. An install that
kept the example ones will now refuse to start, which is the point — see the rotation note in
`docs/self-hosting.md`.

## How to test

Automated:

```bash
bun run typecheck && bun run lint && bun run format:check
cd packages/crypto && bun test                # the validator
bun test scripts/                             # EnvFile.fill and the setup helpers
cd apps/worker && bun test
cd apps/api && bun test --env-file=../../.env.test
```

On the docker stand:

1. `cp .env.example .env`, set `API_URL` / `APP_URL`, leave the secrets empty and run
   `docker compose up -d`: compose refuses with `set POSTGRES_PASSWORD in .env` (the first
   empty guard in file order; each of the five names its own variable) and starts nothing.
2. Put `change-me-please-generate-a-real-secret` into `BETTER_AUTH_SECRET` and
   `WORKER_INTERNAL_TOKEN`, fill the rest with real values, `docker compose up -d`: the api
   container exits at startup with
   `BETTER_AUTH_SECRET is an example value: generate one with \`openssl rand -base64 32\`.`;
   with a real auth secret but the example worker token, the api, the worker and the bot
   each exit with the same message for `WORKER_INTERNAL_TOKEN`.
3. Fill every secret with `openssl rand -base64 32` and `docker compose up -d`: the stack
   boots, sign-up and sign-in work, the worker delivers a notification and the bot fetches
   its config (both go through `/internal/*` with the token).
4. From a checkout: importing `@repo/auth` with an example or short `BETTER_AUTH_SECRET`
   throws before better-auth is constructed; `apps/worker/src/internal-api.ts` and the bot's
   `botConfig()` throw on a short, unset or example token and load with a generated one.
5. `bun run setup` → Develop on a machine without the dev volume: `.env` ends up with
   `POSTGRES_PASSWORD=itsaplan`, the MinIO keys `minioadmin`, and three generated secrets;
   `bun run dev` starts and uploads work.

## Checklist


croffasia:
This way we could silently break instances without telling the user anything. Let's think this through properly :) If we're deciding which secrets they use for them, they need to understand what's going on.

egorvas:
You are right, and the check was worse than it looked. Every refusal is gone. 31 files down to 8.

What I removed and why each one deserved it:

- The startup throw. An instance already running on an example value is not made safer by refusing to serve; the secret was forgeable before and is forgeable after, the instance is just down. And the operator who upgrades is the one person guaranteed to be watching, so a warning reaches them at the same moment a crash would.
- The 32-byte rule. It judged a secret the operator chose privately. Someone running a 24-character random value is not vulnerable, and we would have taken their instance down for length alone. That is the part where we were deciding for them.
- The per-request token check. Worst of the three, and I had not thought it through: it refused internal requests whenever `WORKER_INTERNAL_TOKEN` was weak, **even when the caller presented the correct token**. An instance with a short token would have lost notifications and Telegram quietly. A silent functional break is worse than a loud one.

Then I checked what better-auth 1.6.27 already does, and most of the rest was duplication. `validateSecret` throws on an unset secret, throws in production on its own default, and warns below 32 characters or 120 bits of estimated entropy. The one thing it cannot know about is the value this repository published: `change-me-please-generate-a-real-secret` is 39 characters and scores 144.3 on its own entropy formula, so it clears all three checks in silence. That single string is what is left — four lines in `packages/auth/src/index.ts` that warn and carry on.

What I kept, and where I would push back a little: the empty placeholders in `.env.example`. They cannot break an existing instance, because nobody copies `.env.example` over a live `.env` while upgrading. They reach a new install only, where `${VAR:?}` stops the stack before there is anything to break and the error names both the variable and the file. And they make an existing promise true: `docs/self-hosting.md` already tells the reader "the stack refuses to start while one of these is missing", which is false today.

Documentation alone was the other option and I do not think it holds, because it has already been tried. `SECURITY.md` says "Never reuse the example values" today, and it did not prevent this: the documented quickstart is `cp .env.example .env` plus `docker compose up`, and that hands you a working instance signed with a secret anyone can read from GitHub. While copy-paste produces something that runs, a sentence asking people not to do it loses. `docs/self-hosting.md` now also tells existing operators plainly to replace both values and what happens when they do.

Verification: full api suite 1806 pass / 0 fail, `bun test scripts/` 8 pass, `format:check`, `lint`, `typecheck` clean. Checked the warning by importing the module with each value: it fires on the example string and stays silent on a generated one.

egorvas:
Merged `main` in to clear the conflict #373 created. `WORKER_INTERNAL_TOKEN` is gone from the repository, so its block went with it rather than being kept as an empty placeholder, and the note in `docs/self-hosting.md` now talks about `BETTER_AUTH_SECRET` alone. Against current `main` this is 8 files.

The lint job here is red for a reason that has nothing to do with this branch: `main` itself fails `i18n-json/identical-keys` because `messages/id/apiKeys.json` is missing the five API key expiry strings the other seven locales carry. #363 branched before #358 landed. #381 adds them.

PR #353, opened 2026-09-09
fix(api): bound issue texts and make the mention scan linear

egorvas:
## What

- `POST /issues/:issueId/comments`, `PATCH /comments/:commentId`, the issue `description`
  (create and update) and a string custom field value (`PUT /issues/:issueId/fields/:fieldId`)
  are bounded at 50,000 characters (`ISSUE_TEXT_LIMIT` in `modules/issues/model.ts`, the same
  ceiling the pull-request description already used); a longer text is a 400.
- `parseMentionHandles` scans a text in linear time. The character classes of the link and
  autolink patterns in `NOT_A_MENTION_RE` exclude their own opener, so an unclosed span stops
  at the next `[` or `<` instead of running to the end of the text from every opener. A text
  longer than 100,000 characters (`MAX_MENTION_TEXT_LENGTH`) yields no mentions without being
  scanned.
- `apps/api/src/index.ts` passes `maxRequestBodySize` to `app.listen`, read from
  `API_MAX_REQUEST_BODY_MB` (default 256 MB). Documented in `.env.example` and passed through
  by the three deploy compose files and the Helm chart.

## Why

A comment, description or markdown field value had no upper bound, and the markup scan that
precedes mention parsing backtracked quadratically on a text built from unclosed markup
openers. Measured on the unpatched parser: 27 ms for 10 KB, 0.82 s for 50 KB, 4.1 s for 100 KB
and about 9 s for 200 KB. That runs on the request thread of a single-process Bun server, so
one accepted comment stalled every other request. With the linear patterns the same 100 KB
inputs take under 0.6 ms, and the texts that reach the parser are now bounded at 50,000
characters.

The server also relied on Bun's implicit default for the request body ceiling, and Bun reads a
body in full before any route runs — so an unauthenticated request to `/mcp` or a repository
webhook could make the process buffer an arbitrarily large body before any auth check. The
ceiling is now explicit and configurable. It is one value for the whole server, so it has to
cover the largest attachment an operator allows: the multipart upload is buffered by
`t.File()`, and the import route carries the file as base64 in JSON. `maxAttachmentMb`
(god mode, default 25, settable up to 10,240) must therefore stay below
`API_MAX_REQUEST_BODY_MB`; the `.env.example` comment and the chart values say so.

## How to test

1. `cd apps/api && bun test --env-file=../../.env.test src/shared/__tests__/unit/mentions.test.ts` —
   the linear-time cases finish in under a millisecond; without the `mentions.ts` change they
   take seconds and blow the 50 ms bound.
2. `cd apps/api && bun test --env-file=../../.env.test src/modules/issues/__tests__/integration/activity.test.ts src/modules/issues/__tests__/integration/issues.test.ts` —
   a 50,000-character comment, description and field value are accepted; 50,001 is a 400.
3. Body cap, by hand: `API_PORT=3103 API_MAX_REQUEST_BODY_MB=1 bun --env-file=.env apps/api/src/index.ts`,
   then `curl -X POST http://127.0.0.1:3103/mcp --data-binary @<2 MB file>` answers 413 before
   any handler runs; a 0.5 MB body reaches the route.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,10 @@ BETTER_AUTH_SECRET=change-me-please-generate-a-real-secret  # openssl rand -base
 
 # ── API (Elysia) ──
 API_PORT=3000
+# Largest request body the api accepts, for every route. An attachment upload has to
+# fit under it with its multipart or base64 framing, so keep the god-mode attachment
+# limit (maxAttachmentMb) below this value.
+# API_MAX_REQUEST_BODY_MB=256
 
 # ── Database backup taken before migrations ──
 # The api dumps the database on startup whenever a release has migrations to apply,
--- a/.env.example
+++ b/.env.example
@@ -39,10 +39,6 @@ BETTER_AUTH_SECRET=change-me-please-generate-a-real-secret  # openssl rand -base
 
 # ── API (Elysia) ──
 API_PORT=3000
-# Largest request body the api accepts, for every route. An attachment upload has to
-# fit under it with its multipart or base64 framing, so keep the god-mode attachment
-# limit (maxAttachmentMb) below this value.
-# API_MAX_REQUEST_BODY_MB=256
 
 # ── Database backup taken before migrations ──
 # The api dumps the database on startup whenever a release has migrations to apply,
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -1,8 +1,24 @@
 import { app } from './app';
 
+const MB = 1024 * 1024;
+
+// Bun reads a request body in full before any route runs, and the ceiling is one
+// value for the whole server. An attachment upload has to fit under it with its
+// multipart or base64 framing, so maxAttachmentMb (god mode) must stay below it.
+function maxRequestBodySize(): number {
+  const mb = Number(process.env.API_MAX_REQUEST_BODY_MB || 256);
+  if (!Number.isInteger(mb) || mb <= 0) {
+    throw new Error('API_MAX_REQUEST_BODY_MB must be a positive integer number of megabytes');
+  }
+  return mb * MB;
+}
+
 // Bind the port. The app itself is assembled in ./app.ts (without `.listen()`)
 // so tests can import it and drive routes in memory.
-app.listen(Number(process.env.API_PORT ?? 3000));
+app.listen({
+  port: Number(process.env.API_PORT ?? 3000),
+  maxRequestBodySize: maxRequestBodySize(),
+});
 
 console.log(`🦊 API running at http://${app.server?.hostname}:${app.server?.port}`);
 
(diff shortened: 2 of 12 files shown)


croffasia:
Looks like setting the limit in the env could break the file limits configured in the UI?

egorvas:
You are right, and it turned out worse than a conflict between the two limits: the body ceiling had no reason to be in this pull request at all. Removed it, and retitled accordingly.

Measured on Bun 1.4.0 with a bare `Bun.serve` and no option set: a body of exactly 134217728 bytes (128 MiB) answers 200, one byte more answers 413. Elysia never sets the option itself, it only declares it in its types, so `app.listen(port)` on `main` already runs under that ceiling. The audit item this came from read "no HTTP body limit, Bun default 128 MB", which I took as "no limit". There was one. My change set the default to 256 MiB, so on an instance that never sets the variable it doubled what the server accepts instead of capping it. Exactly the wrong direction.

That also answers your question by removal. There is no longer an environment value that can undercut `maxAttachmentMb` from god mode. Worth recording what I found while checking, in case you want it later: the attachment routes accept both multipart and base64 JSON, and base64 inflates a body by a third, so any future transport ceiling has to clear `maxAttachmentMb` plus a third plus framing, and the two would need to validate against each other. Also `SECURITY.md` puts denial of service through raw request volume out of scope, which is what a body ceiling defends against.

What stays is the part that closes the real attack, and it is not a volume attack at all. A comment of 128 KB, far under any ceiling, made the markup pre-scan backtrack quadratically. Measured here on the two regexes, same input, same machine:

| text | before | after |
| --- | --- | --- |
| 32 KB of `[` | 189 ms | 0.2 ms |
| 64 KB of `[` | 751 ms | 0.3 ms |
| 128 KB of `[` | 2994 ms | 0.5 ms |
| 96 KB of `[` then `@ada` | 1684 ms | 0.4 ms, still resolves `ada` |

Four times the time for twice the input, which is the quadratic signature. The fix is the character classes excluding their own opener plus a length guard, and `maxLength: 50_000` on the description, the comment and a markdown field value, which are the three texts that reach the scan.

One honest trade-off in the regex, verified rather than assumed. A link whose label or url contains a nested `[` is no longer recognised as link markup, so `[see [1] @ada](https://example.com)` and `[@ada](https://example.com/a[b])` now notify `ada`, where before the `@` was masked. Ordinary links, images and autolinks are unaffected. It over-notifies in exotic markdown rather than under-notifies, and anyone who can write that could write a plain mention instead.

Verification: `bun test src/shared/__tests__/unit/mentions.test.ts src/modules/issues` 307 pass / 0 fail; `format:check`, `lint`, `tsc --noEmit` clean. The full api suite ran green on the previous revision of this branch, and this commit only removes files from it.

egorvas:
The failing lint job is not from this branch. `main` fails `bun run lint` on its own right now: `messages/id/apiKeys.json` is missing the five API key expiry strings the other seven locales have, so `i18n-json/identical-keys` errors. #381 fixes it, and this branch goes green behind it.

croffasia (approved):

PR #354, opened 2026-09-09
refactor(api): serve /internal on a private listener, constant-time token

egorvas:
## What

- The five `/internal/*` routes (agent run execution, agent thread cleanup, notification send, Telegram bot config, Telegram link) are served by a second Elysia instance, `internalApp`, which `apps/api/src/index.ts` binds to `INTERNAL_PORT` (default 3002). The public app answers 404 for `/internal/*` and the OpenAPI document no longer lists them.
- `docker-compose.yml`, `docker-compose.coolify.yml`, `docker-compose.coolify-images.yml` and the Helm chart set `INTERNAL_PORT` on the api and point the worker and the bot at it with `SERVICE_URL_API_INTERNAL` (`http://api:3002` in compose, `http://<release>-api:3002` in the chart, exposed on the ClusterIP Service only). No `ports:` mapping and no Ingress route is added for it. `apps/api/Dockerfile` adds `EXPOSE 3002`.
- The worker (`apps/worker/src/internal-api.ts`) and the bot (`apps/bot/src/config.ts`) read `SERVICE_URL_API_INTERNAL`; without it they take the host of the api origin (`SERVICE_URL_API`, else `API_URL`) on `INTERNAL_PORT`, which is `http://localhost:3002` in local dev. The bot only ever called `/internal/*`, so its base URL is the internal one.
- The three `x-worker-token` comparisons use one helper, `apps/api/src/shared/worker-token.ts`, built on `crypto.timingSafeEqual` with a length pre-check, mirroring `modules/git/providers.ts` and `packages/auth/src/instance.ts`.
- `POST /internal/notification-deliveries/send` accepts `{ id }` of a pending `notification_delivery` row and reads channel, recipient, project and payload from that row (`getPendingDelivery` in `notifications/outbound.ts`). An unknown or non-pending id is a 400. The worker's claim query returns only `id` and `attempts` and the send call passes the id; the claim lease, retry and failure paths are unchanged.
- `internalApp` carries its own `onError`, so `HttpError`, validation failures and unexpected errors answer `{ error }` JSON on that listener the way the planner does; the worker already reads `body.error`.
- Docs: `.env.example` documents `INTERNAL_PORT` and `SERVICE_URL_API_INTERNAL`; `docs/self-hosting.md`, `docs/coolify.md`, the chart README and the `AGENTS.md` of api, worker and bot say the port must stay private.

## Why

The internal routes were mounted on the same listener as the public API, so every self-hosted stack published them on the api's public port. They were protected by a shared token compared with `===`, and one of them accepted a caller-chosen recipient and message. Together this exposed the class of issue where a reachable internal endpoint with a guessable or timing-leakable credential gives access to instance secrets, arbitrary mail sending and agent execution. Moving the routes onto a port that only the compose network or the cluster reaches, comparing the token in constant time, and making the mail route send only what the outbox already holds closes the three parts of that.

Closes C2 (internal audit).

## How to test

Automated (from `apps/api`, with `.env.test`):

```
bun test --env-file=../../.env.test src/__tests__/integration/internal-listener.test.ts
bun test --env-file=../../.env.test src/__tests__/integration/openapi.test.ts
bun test --env-file=../../.env.test src/modules/invites/__tests__/integration/invites.test.ts
bun test --env-file=../../.env.test src/modules/agents/core/__tests__/integration/run-limits.test.ts
```

`internal-listener.test.ts` checks that the public app answers 404 for every `/internal/*` route even with the right token and lists none of them in `/docs/json`; that the internal instance answers 401 with no token, a wrong token of another length, a wrong token of the same length, and when the instance has no token configured, and 200 with the right one; and that the send route rejects a body carrying a recipient and message (400), rejects an unknown id (400), and sends the row a real id names.

On the docker stand:

1. `docker compose up -d --build`, then `docker compose port api 3002` prints nothing and `curl -H 'x-worker-token: <WORKER_INTERNAL_TOKEN>' http://<host>:3000/internal/telegram/config` answers 404. The api log shows both `API running at ...:3000` and `Internal API running at ...:3002`.
2. From inside the network, `docker compose exec worker wget -qO- --header "x-worker-token: $WORKER_INTERNAL_TOKEN" http://api:3002/internal/telegram/config` answers `{"enabled":false,...}` (or the configured bot).
3. Worker delivery over the internal port: configure an email provider in god mode, invite an address to a project (the invite response has `emailQueued: true`), and watch the worker log: the row is claimed and the api answers the send; the `notification_delivery` row disappears on success or carries `last_error` on a provider failure. No worker log line mentions `SERVICE_URL_API or API_URL is required`.
4. Bot over the internal port: set a bot token in god mode; the bot log shows it read the settings and started polling, and `/start <code>` from a profile link completes the link.
5. Agent runs: mention an internal agent on an issue; the worker claims the run and the api executes it over `http://api:3002/internal/agent-runs/execute` (the run row reaches `succeeded` or `failed` with the model's error, not a connection error).

## Checklist

- [ ] Database schema changed: migration generated with `bun run db:generate` and committed (no schema change)

> Overlaps #352 (`fix/auth-fail-closed-on-weak-secrets`), which adds `apps/api/src/shared/worker-token.ts` for the secret-strength check. That PR is meant to land first; rebasing this one on it keeps that file and replaces its `===` with the constant-time comparison here.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,9 @@ BETTER_AUTH_SECRET=change-me-please-generate-a-real-secret  # openssl rand -base
 
 # ── API (Elysia) ──
 API_PORT=3000
+# Second listener serving the /internal/* routes the worker and the bot call. Keep
+# it off the public network: the compose files and the chart never publish it.
+INTERNAL_PORT=3002
 
 # ── Database backup taken before migrations ──
 # The api dumps the database on startup whenever a release has migrations to apply,
@@ -71,6 +74,10 @@ APP_ENCRYPTION_KEY=
 # ── Internal token shared by api, worker and bot ──
 # Same value for all three. openssl rand -base64 32
 WORKER_INTERNAL_TOKEN=change-me-please-generate-a-real-secret
+# Where the worker and the bot reach the api's internal listener. Inside compose it
+# is http://api:3002 (set by the compose files); unset, both derive it from the
+# API_URL host and INTERNAL_PORT, which is http://localhost:3002 in local dev.
+# SERVICE_URL_API_INTERNAL=http://localhost:3002
 
 # ── Outbound request policy (apps/api + apps/worker) ──
 # Server-side fetches of a URL that came from a request — a webhook target, an
--- a/apps/api/AGENTS.md
+++ b/apps/api/AGENTS.md
@@ -15,7 +15,10 @@ Rules and invariants for this package below; read the code for the walkthrough.
   `model.ts` and is re-exported from each child's (`agentParams`).
 - `src/app.ts` assembles and exports the app (`export const app`, no `.listen()`);
   `src/index.ts` only binds the port. `export type App = typeof app` types the Eden
-  Treaty client (web + tests).
+  Treaty client (web + tests). The `/internal/*` routes the worker and the bot call
+  live on a second instance, `internalApp`, which `index.ts` binds to `INTERNAL_PORT`
+  (3002); the public listener never serves them. A new internal route goes on that
+  instance and checks the token with `#shared/worker-token`.
 - `index.ts`: `new Elysia({ name: "<feature>", detail: { tags: ["<Tag>"] } })` —
   routes chained directly on it, each route sets `detail.summary`. Handlers only;
   the schemas they reference come from `model.ts`.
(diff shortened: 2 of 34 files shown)


croffasia:
@egorvas  Thanks for flagging this. We'll rework it differently - dropping the internal routes and the worker token.

croffasia:
Fixed #373

PR #355, opened 2026-09-09
fix(api): skip delegation and field triggers started by an agent

egorvas:
## What

`enqueueDelegateRun` and `enqueueFieldRun` in `apps/api/src/modules/issues/service.ts`
return early when the acting user is an agent's bot user (`isAgentUser`), the same
check the mention trigger applies to a comment's author. The guard covers every way an
agent acts on an issue: the internal runtime's tool dispatch and the MCP endpoint both
authenticate with the agent's API key, which resolves to the agent's bot user id and is
what the route hands to the service as the actor.

## Why

An agent-authored delegation or member-field change could queue a run for another
agent (or, through a second agent, for itself), and that run could do the same in
return. Two agents wired this way run each other indefinitely, and each run is a billed
LLM call with no spend cap on a self-hosted instance. The mention path already has this
loop guard; this brings the other two triggers in line with it.

## How to test

1. Create two internal agents in a project, A and B, with B's "runs on delegation"
   turned on.
2. Have A delegate an issue to B (A acting through its API key, e.g. over MCP):
   B's run history stays empty.
3. Delegate the same issue to B as a person: B gets a delegation run.
4. Repeat 2 and 3 with a member custom field that B carries a field trigger for.

Automated: `cd apps/api && bun test --env-file=../../.env.test src/modules/agents/core/__tests__/integration/agent-runs.test.ts`
(the two new `loop guard` cases fail without the fix and pass with it).

## Checklist

### DIFF
--- a/apps/api/src/modules/agents/core/__tests__/integration/agent-runs.test.ts
+++ b/apps/api/src/modules/agents/core/__tests__/integration/agent-runs.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach } from 'bun:test';
-import { authedApi, type Api } from '#tests/helpers/app';
+import { apiKeyApi, authedApi, type Api } from '#tests/helpers/app';
 import { signUpTestUser } from '#tests/helpers/auth';
 import { resetDb } from '#tests/helpers/db';
 import { createAgent, projectIdOf, teamOf } from '#tests/helpers/agents';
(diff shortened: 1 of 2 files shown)


croffasia:
Delegating tasks between agents was part of the design from the start. Closing this PR - if you have ideas for improving it, happy to discuss in Discussions.

PR #357, opened 2026-09-09
fix(runner): keep the coding CLI's approval gate on unless the operator opts out

egorvas:
## What

- `packages/runner` presets no longer pass the flags that turn a CLI's approval gate
  off. `--permission-mode auto` (Claude Code), `--dangerously-skip-permissions`
  (Antigravity CLI) and `--allow-all-tools` (Copilot CLI) moved into a new `skipApprovals`
  field on `Preset`, added to the argv only when the operator opts in.
- New runner setting `skipApprovals`: `"skipApprovals": true` in `itsaplan-runner.json`,
  `ITSAPLAN_SKIP_APPROVALS=1`, or `--skip-approvals` on the command line. Default `false`.
  It follows the file/env/CLI/agent-entry precedence the other settings already use, and a
  non-boolean value is rejected at load with `skipApprovals must be true or false`. The
  startup line says when it is on.
- `--no-ask-user` stays in the Copilot preset unconditionally: it turns off the tool that
  would block on a person's answer, it grants nothing. Codex and opencode gain no flags —
  `codex exec` runs in its own sandbox (`sandbox_mode="workspace-write"`) and opencode
  takes permissions from its own config.
- The agent system prompt now says which part of what it reads is text from the tracker.
  `runModePreamble()` — used by both agent paths, the internal one in
  `agents/core/internal-routes.ts` and the runner one in `agents/runner/service.ts` —
  gained an "Untrusted text" block, and `framePrompt()` wraps the trigger comment, the
  thread above it and a schedule's task in `<user_text>` delimiters. A delimiter written
  inside that text is neutralised, so it cannot close the block and continue as framing.
- `docs/runner.md` gained a "Threat model" section, and both it and
  `packages/runner/README.md` document the flag and the changed default.

## Why

A run's task is text from the tracker: the comment that mentioned the agent, the thread
above it, an issue body, a schedule's prompt. Anyone who can comment in the project writes
part of it. The external runner feeds that text to a coding agent CLI on the operator's own
machine, with the operator's files and environment, and the shipped presets removed the one
thing standing between the two — the CLI's approval gate. A prompt-injection class of issue
therefore reached the operator's shell without anyone opting into that risk. Only operators
running `@itsaplan/runner` were affected; internal agents run in the worker and never
spawn a CLI.

Defaulting to the gate is the fix: a tool call the CLI would ask a person about is denied,
the run continues with what it could do, and the transcript shows what was denied. An
operator who wants an unattended machine can still have one, deliberately, after reading
the threat model.

## How to test

1. `cd packages/runner && bun test` — the preset test asserts that no preset emits an
   approval-skipping flag by default, and that each one emits its own with the opt-in.
2. Run a runner against an instance with `--agent claude` and no `--skip-approvals`:
   the startup line names no approvals, and `claude` is spawned without
   `--permission-mode`. Add `--skip-approvals` (or `ITSAPLAN_SKIP_APPROVALS=1`, or
   `"skipApprovals": true` in the config file) and `--permission-mode auto` is back.
3. `cd apps/api && bun test --env-file=../../.env.test src/modules/agents` — a claimed
   run's `systemPrompt` carries the untrusted-text block and its `prompt` carries the
   comment inside `<user_text>`.
4. Mention an agent in a comment and read the run's prompt: the comment and the thread
   above it are inside the delimiters, the surrounding instructions are not.

## Checklist

`.env.example` is the instance's environment. `ITSAPLAN_SKIP_APPROVALS` is read by the
runner on the operator's own machine, like every other `ITSAPLAN_*` variable, none of which
appear there; it is documented in the runner's README settings table and in `docs/runner.md`.

### DIFF
--- a/apps/api/src/modules/agents/core/__tests__/unit/framing.test.ts
+++ b/apps/api/src/modules/agents/core/__tests__/unit/framing.test.ts
@@ -1,6 +1,71 @@
 import { describe, it, expect } from 'bun:test';
-import { projectPreamble } from '../../prompt/framing';
+import {
+  asUserText,
+  framePrompt,
+  projectPreamble,
+  runModePreamble,
+  USER_TEXT_TAG,
+  type RunForPrompt,
+} from '../../prompt/framing';
 import { PROJECT_DESCRIPTION_LIMIT } from '#modules/projects/model';
+
+const OPEN = `<${USER_TEXT_TAG}>`;
+const CLOSE = `</${USER_TEXT_TAG}>`;
+
+const mention: RunForPrompt = {
+  id: 1,
+  trigger: 'mention',
+  prompt: 'please review @bot',
+  issueId: 7,
+  issueIdentifier: 'MKT-7',
+  issueTitle: 'Landing page',
+  assigneeName: null,
+  requesterName: 'Ada',
+  requesterUsername: 'ada',
+  agentUserId: 'u-bot',
+  agentUsername: 'bot',
+  threadContext: 'Ada: first draft is up',
+  sourceActivityId: 3,
+};
+
+// A run is started by text anyone in the project can write, and the same text is what
+// the agent is asked to act on, so the system prompt has to say which is which and the
+// task has to mark where that text is: without either, a comment can be read as the
+// framing around it.
+describe('runModePreamble', () => {
+  it('tells every kind of run to read user text and tool results as data', () => {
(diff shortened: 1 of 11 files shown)


croffasia:
This isn't a bug! You get full control over how the external runner is configured, including how the agent is launched under it. Turning off auto mode would mean a hundred questions before anything actually gets done.

PR #361, opened 2026-09-09
fix(api): vet the team SMTP host against the SSRF guard

egorvas:
## What

- `PUT /teams/:teamId/notification-settings` resolves the SMTP host and rejects it
  with 400 when it points to a loopback, link-local, or private address (or does not
  resolve), unless the host is named in `SSRF_ALLOWED_HOSTS`. The check runs only
  when the SMTP section is enabled, so a leftover host does not block switching to
  Resend or the instance provider.
- `smtp.timeout` is bounded to 1..60 seconds (`SMTP_TIMEOUT_MAX_SECONDS`).
- The default SMTP encryption for a team that has saved nothing is `tls` (STARTTLS,
  matching the default port 587). `none` stays selectable.
- `@repo/net` gains `assertPublicHost(host, label?)`: the resolve-and-check part of
  `assertPublicHttpUrl`, for a supplied host that is connected to by a non-http
  client. `apps/api/src/shared/net.ts` wraps it into the API's 400.
- `.env.example` documents that `SSRF_ALLOWED_HOSTS` also governs the SMTP host, so
  an internal mail relay on a private address can still be named there.

## Why

The team SMTP host was stored with only trim and non-empty validation, while the
delivery sender (`/internal/notification-deliveries/send`, running in the API
process) connects to it on every email. Any team owner could point it at an
internal address and port and use the delivery outcome as a blind connectivity
probe. Webhook targets, attachment imports, and repository hosts already go
through the `@repo/net` guard; the SMTP host now does too. The unbounded timeout
let a single delivery hold the API request for as long as the caller chose, and
`none` as the default encryption sent credentials in clear to whatever answered.

## How to test

1. As a team owner, open the team's notification settings and pick SMTP.
2. Save with host `127.0.0.1`, `10.0.0.1`, `169.254.169.254`, or `localhost`: the
   save fails with "SMTP host must not point to a private or local address".
3. Save with a public relay hostname: the save succeeds.
4. Save with a timeout of 61: rejected; 60: accepted.
5. Set `SSRF_ALLOWED_HOSTS=10.0.0.1` on the api, restart, repeat step 2 with
   `10.0.0.1`: accepted.
6. A team with no saved settings shows encryption "TLS" and port 587 in the form.
7. Resend settings are unaffected: saving a Resend key works as before.

Automated: `cd apps/api && bun test --env-file=../../.env.test src/modules/notification-settings`
and `cd packages/net && bun test`.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,26 +73,29 @@ APP_ENCRYPTION_KEY=
 WORKER_INTERNAL_TOKEN=change-me-please-generate-a-real-secret
 
 # ── Outbound request policy (apps/api + apps/worker) ──
-# Server-side fetches of a URL that came from a request — a webhook target, an
-# attachment to import, a repository host — are refused when the URL resolves to a
-# loopback, link-local, or private address. Name a host here to trust it anyway.
+# Server-side connections to a host that came from a request — a webhook target, an
+# attachment to import, a repository host, a team's SMTP relay — are refused when
+# the host resolves to a loopback, link-local, or private address. Name a host here
+# to trust it anyway.
 #
-# The case this exists for is a self-hosted repository host: its address is
-# configuration rather than content, and on a self-hosted instance it is normally
-# private, so without this a Gitea/Forgejo/GitLab on your own network cannot be
-# connected at all. Exact hostnames, comma-separated; no wildcards or CIDR ranges.
-# https is still required and the resolved address is still pinned.
+# The cases this exists for are a self-hosted repository host and a mail relay on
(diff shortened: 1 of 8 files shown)


croffasia:
Closing this. Only the worker connects to the SMTP host, delivery is async and we expose no delivery log, so there is nothing to read a scan from. The check would also 400 self-hosted instances whose relay is a compose service or a LAN address. The timeout bound and the tls default are worth having on their own - happy to take those separately.

PR #390, opened 2026-09-13
fix: restore MinIO image pulls through Quay

OthmanAdi:
Fresh installs and CI fail before startup because Docker Hub denies pulls for `minio/minio:latest` and `minio/mc:latest`. This affects [upstream main](https://github.com/croffasia/itsaplan/actions/runs/34653924713) and [the current navigation PR](https://github.com/croffasia/itsaplan/actions/runs/34753580607).

Use MinIO's official Quay registry for both images across the self-hosting, development, test, and two Coolify Compose files. The ten substitutions preserve the tags, commands, and storage configuration. MinIO's [own chart](https://github.com/minio/minio/blob/7aac2a2c5b7c882e68c1ce017d8256be2feea27f/helm/minio/values.yaml#L16) uses these repositories; its [client publishing script](https://github.com/minio/mc/blob/77f82e18b5401a65958f1619df6ebb994634bd88/docker-buildx.sh#L25) publishes the same build to both registries.

Validation: both Quay images pulled successfully. Their current manifest digests exactly match the cached Docker Hub images, with Linux amd64, arm64, and ppc64le available. Root formatting, lint, and typecheck passed, as did tidy and code review. Compose validation passed for self-hosting, development, and tests. The Coolify files passed with their existing `exclude_from_hc` metadata omitted only from validation input, since standard Compose rejects that Coolify-specific field on unchanged upstream too. The same MinIO image contents passed the local Docker gate with 1,915 tests.


croffasia (approved):

shiouen:
@croffasia @OthmanAdi The helm chart needs an update too, I'll set up an issue

PR #393, opened 2026-09-13
improvement: load visible mobile link previews with backend caching

OthmanAdi:
## Summary

Compact link cards load preview metadata automatically when they enter the visible area. Offscreen cards wait until they are visible. Tapping the link opens its destination; the separate preview button opens a mobile dialog with Open, Copy and Edit actions. Desktop hover previews remain available.

The API caches public metadata for five minutes and empty results for 15 seconds, deduplicates concurrent requests, and bounds entries and retained bytes. Every request still validates its URL and public DNS resolution before cache lookup. Existing pinned-fetch, redirect and image protections remain in place. Internal previews remain scoped to the active session and recheck access.

Includes current upstream changes through `2aa58ec6`.

## Validation

- Root format, lint, typecheck and app build passed.
- 309 web tests, 7 root script tests, 18 focused cache/metadata tests passed.
- Docker integration gate: 1,942 API, 4 bot and 79 worker tests passed.
- Mounted regression tests cover visible-only requests, cleanup, document visibility, internal access refresh, URL normalization, cache expiry and eviction, concurrent requests and rejection recovery.
- Chrome touch emulation at 390 × 844: only the two visible cards requested metadata; three more loaded after scrolling into view. Opening an already loaded preview made no additional metadata request. No browser console errors were observed.
- Physical iPhone Safari verification remains outstanding.

## Screenshots

Fictional acceptance data, captured from the integrated build containing this change.

### Automatically loaded visible cards

Visible mobile link cards

### Expanded preview

Expanded mobile link preview


croffasia:
Better to go with the standard approach: load previews upfront with caching on the backend, and only for the cards that are actually visible to the user. That's the better UI/UX solution.

image
image

croffasia (approved):

